PR #47, opened 2016-12-20
Adding AppVeyor CI

jtaubensee:
Currently fails when running the tests from PowerShell. Will investigate.
2016-12-20_14h58_52


jtaubensee:
No longer receiving the above error.

jtaubensee:
@SeanFeldman - Regarding the unit tests? Yes, unfortunately that is true. We will need to think about implementing a private build that can be run without risk of exposing the credentials.

SeanFeldman:
Right. Reason I'm asking is that there's no way to know if PR is failing or
not and it would require manual verification on each PR.

jtaubensee:
Yep, I totally agree. As we have more and more PR's my team and I will need to do this for efficiency purposes. Stay tuned...

PR #51, opened 2016-12-22
Fixing environment variable name in exception

SeanFeldman:
When environment variable is missing, wrong name is shown in exception.


SeanFeldman:
@azure/azure-service-bus-write please review

PR #78, opened 2017-01-27
Provide a way to determine if message was received

### LINKED ISSUE #68: Extend API for BrokeredMessage: IsReceived bool property
The BrokeredMessage API is problematic in the sense that there are several public properties that may throw exceptions, when attempting to read them. This includes all the properties that are only available for a BrokeredMessage that has been sent by Service Bus.

## Scenario ##

I am currently attempting to write a generic WebAPI MediaTypeFormatter for BrokeredMessage. With the current API, this leads to ugly code:
```csharp
public static bool IsReceived(BrokeredMessage message)
{
    try
    {
        int dummy = message.DeliveryCount;
        return true;
    }
    catch (InvalidOperationException)
    {
        return false;
    }
}
```

## API proposal ##
I propose to add a new property:
```csharp
bool IsReceived { get; }
```

The definition would be: `IsReceived` will return `true` if and only if `ThrowIfNotReceived` does not throw. Or by the current implementation, if `receiverHeaders != null`.

SeanFeldman:
Fixes #68

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
@@ -130,6 +130,12 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                     await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
                     var receivedMessages = await TestUtility.ReceiveMessagesAsync(queueClient.InnerReceiver, 1);
                     Assert.True(receivedMessages.First().IsReceived);
+
+                    // TODO: remove when per test cleanup is possible
+                    if (receiveMode == ReceiveMode.PeekLock)
+                    {
+                        await receivedMessages.First().CompleteAsync();
+                    }
                 }
                 finally
                 {
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
@@ -4,6 +4,8 @@
 namespace Microsoft.Azure.ServiceBus.UnitTests
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using System.Threading.Tasks;
     using Xunit;
 
@@ -104,6 +106,36 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
             Assert.Equal("id-2", message2.MessageId);
 
             BrokeredMessage.SetMessageIdGenerator(null);
+        }
+
+        [Fact]
+        [DisplayTestMethodName]
+        void Should_return_false_for_message_that_was_not_sent_and_received()
+        {
+            var message = new BrokeredMessage();
+            Assert.False(message.IsReceived);
+        }
+
+        [Theory]
+        [DisplayTestMethodName]
+        [InlineData(ReceiveMode.ReceiveAndDelete)]
+        [InlineData(ReceiveMode.PeekLock)]
+        async Task Should_return_true_for_message_that_was_sent_and_received(ReceiveMode receiveMode)
+        {
+            var message = new BrokeredMessage();
+            Assert.False(message.IsReceived);
+
+            var queueClient = QueueClient.CreateFromConnectionString(TestUtility.GetEntityConnectionString(Constants.NonPartitionedQueueName), receiveMode);
+            try
+            {
+                await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
+                var receivedMessages = await TestUtility.ReceiveMessagesAsync(queueClient.InnerReceiver, 1);
+                Assert.True(receivedMessages.First().IsReceived);
+            }
+            finally
+            {
+                await queueClient.CloseAsync();
+            }
         }
     }
 }
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
@@ -162,26 +162,6 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                     await queueClient.CloseAsync();
                 }
             }
-
-            [Fact]
-            [DisplayTestMethodName]
-            async Task Should_return_true_for_completed_message()
-            {
-                var queueClient = QueueClient.CreateFromConnectionString(TestUtility.GetEntityConnectionString(Constants.NonPartitionedQueueName), ReceiveMode.PeekLock);
-                try
-                {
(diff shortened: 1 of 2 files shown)


SeanFeldman:
Ready for review `@group??` (@jtaubensee @binzywu @nemakam @vinaysurya) 😺

vinaysurya:
:shipit:

SeanFeldman:
Please hold. A test for peek case it's missing. I'll added later today.

SeanFeldman:
Done. I've added some context (`When_querying_IsReceived_property`) and covered `Peek` scenario.

test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
```diff
@@ -105,5 +106,55 @@ void CustomMessageIdGenerator()
...
+                [DisplayTestMethodName]
+            async Task Should_return_true_for_peeked_message()
+            {
+                var queueClient = QueueClient.CreateFromConnectionString(TestUtility.GetEntityConnectionString(Constants.NonPartitionedQueueName), ReceiveMode.PeekLock);
+                try
+                {
+                    await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
+                    var peekedMessage = await TestUtility.PeekMessageAsync(queueClient.InnerReceiver);
+                    var result = peekedMessage.IsReceived;
+                    await TestUtility.CompleteMessagesAsync(queueClient.InnerReceiver, new[] { peekedMessage });
```
vinaysurya: >TestUtility.CompleteMessagesAsync(queueClient.InnerReceiver, new[] { peekedMessage }); [](start = 26, length = 86)

wait this works? somethings off,  the complete on a peek message should not work.

SeanFeldman: Looks like I [found a bug](https://github.com/Azure/azure-service-bus-dotnet/pull/81) 😉 
Unit testing FTW.

SeanFeldman: @vinaysurya I'll need to rebase this on top of the updated `dev` once #81 is merged.

Oh, and what are these cryptic comments?
image

vinaysurya: Sorry, that was not intentional.. not sure where that came from

SeanFeldman: I'll blame @jtaubensee for that 😉 

So should I wait for #81 to be merged first?

vinaysurya: yes, I think you might want to change the code to receive the message and complete it also right?

SeanFeldman: Thinking of it, this test should be changed to test for `false` result _and_ an additional test should be created for testing the property returns `true` after message is completed.

SeanFeldman:
@vinaysurya @jtaubensee I'm done. Rebased on top of `dev` to include fix #81 and added one more test case.

test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
```diff
@@ -105,5 +106,76 @@ void CustomMessageIdGenerator()
...
+            [DisplayTestMethodName]
+            [InlineData(ReceiveMode.ReceiveAndDelete)]
+            [InlineData(ReceiveMode.PeekLock)]
+            async Task Should_return_true_for_message_that_was_sent_and_received(ReceiveMode receiveMode)
+            {
+                var queueClient = QueueClient.CreateFromConnectionString(TestUtility.GetEntityConnectionString(Constants.NonPartitionedQueueName), receiveMode);
+                try
+                {
+                    await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
+                    var receivedMessages = await TestUtility.ReceiveMessagesAsync(queueClient.InnerReceiver, 1);
```
vinaysurya: >ReceiveMessagesAsync [](start = 61, length = 20)

For PeekLockMode test, you may have to complete here, so as to not leave behind messages that are picked up by other tests, until we have a pertest purging of the queues

SeanFeldman:
@vinaysurya comments addressed. Please review.

test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
```diff
@@ -105,5 +106,82 @@ void CustomMessageIdGenerator()
...
+            async Task Should_return_true_for_completed_message()
+            {
+                var queueClient = QueueClient.CreateFromConnectionString(TestUtility.GetEntityConnectionString(Constants.NonPartitionedQueueName), ReceiveMode.PeekLock);
+                try
+                {
+                    await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
+                    var peekedMessage = await TestUtility.PeekMessageAsync(queueClient.InnerReceiver);
+                    var messages = await TestUtility.ReceiveMessagesAsync(queueClient.InnerReceiver, 1);
+                    await TestUtility.CompleteMessagesAsync(queueClient.InnerReceiver, messages);
+                    var result = peekedMessage.IsReceived;
```
vinaysurya: >peekedMessage.IsReceived; [](start = 33, length = 25)

We are not really checking the IsReceived on the completed message, still checking on the PeekedMessage, which is similar to test case above... So may be we can take this test of altogether.

SeanFeldman: @vinaysurya if assertion is made on `messages.First.IsReceived` does that fix it then?

vinaysurya: @SeanFeldman, the thing is complete call does not really do any modification on the message after it was received.. It just takes lockToken on the message and sends it to Service to complete the message. So there isnt really a validation necessary?

SeanFeldman: Ok, I'll remove it then. Thanks for clarification 😃

SeanFeldman: @vinaysurya done. please review

vinaysurya:
Thanks Sean! Looks good to me..

At this point, still not very decided if PeekMessages need to be considered as ReceivedMessages. But at this point we will go ahead with this behavior and if more people are interested in using this property in different ways and find that an issue, we might have to re-consider that.

SeanFeldman:
Agree @vinaysurya. That is a very pragmatic approach.

PR #81, opened 2017-01-31
Complete should throw on peeked messages

vinaysurya:
## Description

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
@@ -19,10 +19,6 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                 ReceiveMode.ReceiveAndDelete);
 
             await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
-            var message = await queueClient.PeekAsync();
-            Assert.NotNull(message);
-            await Assert.ThrowsAsync<InvalidOperationException>(async () => await message.CompleteAsync());
-
             message = await queueClient.ReceiveAsync();
             Assert.NotNull((object)message);
 
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                 ReceiveMode.ReceiveAndDelete);
 
             await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
-            message = await queueClient.ReceiveAsync();
+            var message = await queueClient.ReceiveAsync();
             Assert.NotNull((object)message);
 
             await Assert.ThrowsAsync<InvalidOperationException>(async () => await message.CompleteAsync());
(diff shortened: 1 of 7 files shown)


test/Microsoft.Azure.ServiceBus.UnitTests/BrokeredMessageTests.cs
```diff
@@ -19,7 +19,11 @@ public class BrokeredMessageTests
                 ReceiveMode.ReceiveAndDelete);
 
             await TestUtility.SendMessagesAsync(queueClient.InnerSender, 1);
-            var message = await queueClient.ReceiveAsync();
+            var message = await queueClient.PeekAsync();
```
SeanFeldman: This is awesome. Really. Functionality backed up by a test.
But can I ask a favour (as a payback for finding the bug)? Could you please (nice please) extract it into a method that would be self contained and disclose what it does? _Peeked_message_should_not_be_allowed_to_be_completed_ as an example. Bam! Right there if it fails, I know what it does 😃

vinaysurya: Agree, since the tests were already taking time, I tried to reduce the time for another sent message and another receive etc by including that in there. :-)... But agree for clarity this is better, I've seperated that out and published a new pull request.

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
```diff
@@ -41,39 +47,68 @@ public class ExpectedMessagingExceptionTests
...
             }
+        }
 
-            messageSession = await queueClient.AcceptMessageSessionAsync(sessionId);
-            Assert.NotNull((object)messageSession);
-
-            message = await messageSession.ReceiveAsync();
-            TestUtility.Log($"Received Message: SessionId: {messageSession.SessionId}");
+        [Fact]
+        async Task CompleteOnPeekedMessagesShouldThrowTest()
```
SeanFeldman: Just curious - why the `Test` suffix? This code is not going to be executed in production and it's attributed as a test (`[Fact]`).

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
```diff
@@ -18,21 +17,28 @@ public class ExpectedMessagingExceptionTests
...
+                await TestUtility.SendMessagesAsync(queueClient.InnerSender, messageCount);
+                var receivedMessages = await TestUtility.ReceiveMessagesAsync(queueClient.InnerReceiver, messageCount);
 
-            Assert.True(receivedMessages.Count() == messageCount);
+                Assert.True(receivedMessages.Count() == messageCount);
 
-            // Let the messages expire
-            await Task.Delay(TimeSpan.FromMinutes(1));
+                // Let the messages expire
+                await Task.Delay(TimeSpan.FromMinutes(1));
```
SeanFeldman: This is where not having management operations hurts @jtaubensee

A better scenario would be to create an entity with a small lock duration, so that we can continue testing right away.

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
vinaysurya: Just practice i guess, i'll remember to avoid it next time

jtaubensee: We went with that convention to easily identify the tests inside the test class itself (i.e. its not a shared method).

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
jtaubensee: Well based on the way we currently do things, we can create a dedicated entity for this test. It shouldn't be a problem to add to the ARM template, when we decide to do so.

PR #82, opened 2017-01-31
Implemented interfaces for easier testing, and added ServiceBusClientFactory

nemakam:
## Description
Introduced the `ServiceBusClientFactory` to create new client entities and removed client specific static methods. Exposed all other entity clients as interfaces.

1. Exposed `MessageReceiver` and `MessageSender` as public entities which can be operated independently from the Entity clients. (#57 )
1. To make the client test-friendly (#33), all the public methods need to be exposed through interfaces (which can be mocked). As part of this, I've exposed the following: 
    * `IMessageReceiver`
    * `IMessageSender`
    * `IMessageSession`
    * `IMessageSessionEntity`
    * `IQueueClient`
    * `ISubscriptionClient`
    * `ITopicClient`
1. Creation of all entities is now delegated to `ServiceBusClientFactory` which has implementations of all overloads for creating each entity. All the static create methods (e.g., `QueueClient.Create()`) have been removed. **This is a deviation from the original .NET Framework client, by letting users manage the connection objects explicitly rather than having the previous `MessagingFactory` manage them implicitly. The general goal of this library is to be more explicit/straight forward.**
1. `ServiceBusClientFactory` itself can be tested since it has been exposed as `IServiceBusFactory`. All the Create methods now return interfaces instead of the classes. For eg, `CreateQueueClient()` should return `IQueueClient` instead of `QueueClient`. 
1. As part of having uniform interfaces, `IMessageReceiver` and `IMessageSender` are made standard and a few API names have been updated in the clients (Queue/Sub/TopicClient). 

This checklist is used to make sure that common guidelines for a pull request are followed.

- [x] **I have read the [contribution guidelines](./CONTRIBUTING.md).**
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.
- [x] The pull request does not introduce breaking changes (unless a major version change occurs in the assembly and module).
- [ ] If applicable, the public code is properly documented.
- [x] Pull request includes test coverage for the included changes.
- [x] The code builds without any errors.

Issue #33, opened 2016-12-13
Abstracting common components for easier testing


SeanFeldman:
There are scenarios where it would help to be able to test code that relies on ASB client w/o exercising ASB client itself. A few scenarios are:
- Broker not available
- Emulating certain operations to throw exceptions they can raise, but not on demand

With previous client the only option is to wrap everything from ASB to allow mocks/stubs. It would help to have new client to take testability into consideration.

nemakam:
You can track the changes in PR #82

jtaubensee:
This was addressed by #95

PR #85, opened 2017-02-01
Fixes proposed for PR #82

SeanFeldman:
TL&DR: yes, read and checked.

Improvements suggested for PR #82 that addresses issue #33 

@nemakam I might missed a few things, sorry 😊

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
@@ -35,6 +35,30 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
         [Theory]
         [MemberData(nameof(TestPermutations))]
         [DisplayTestMethodName]
+        async Task MessageReceiverAndMessageSenderCreationWorksAsExpected(string queueName, int messageCount = 10)
+        {
+            var messagingFactory = new MessagingFactory();
+            var messageReceiver =
+                (MessageReceiver)messagingFactory.CreateMessageReceiverFromConnectionString(
+                    TestUtility.GetEntityConnectionString(queueName));
+            var messageSender =
+                (MessageSender)messagingFactory.CreateMessageSenderFromConnectionString(
+                    TestUtility.GetEntityConnectionString(queueName));
+
+            try
+            {
+                await this.PeekLockTestCase(messageSender, messageReceiver, messageCount);
+            }
+            finally
+            {
+                await messageSender.CloseAsync();
+                await messageReceiver.CloseAsync();
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(TestPermutations))]
+        [DisplayTestMethodName]
         async Task ReceiveDeleteTest(string queueName, int messageCount = 10)
         {
             var messagingFactory = new MessagingFactory();
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
             var queueClient = (QueueClient)messagingFactory.CreateQueueClientFromConnectionString(TestUtility.GetEntityConnectionString(queueName));
             try
             {
-                await this.PeekLockTestCase(queueClient.InnerSender, queueClient.InnerReceiver, messageCount);
+                await this.PeekLockTestCase(queueClient, queueClient, messageCount);
             }
             finally
             {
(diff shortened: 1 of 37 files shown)


test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
```diff
@@ -174,7 +174,7 @@ public sealed class QueueClientTests : SenderReceiverClientTestBase
             var queueClient = (QueueClient)messagingFactory.CreateQueueClientFromConnectionString(TestUtility.GetEntityConnectionString(queueName), ReceiveMode.ReceiveAndDelete);
```
nemakam: (QueueClient) cast can be removed for all the tests and the object can be directly passed everywhere instead of InnerReceiver or InnerSender

test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs
```diff
@@ -174,7 +174,7 @@ public sealed class QueueClientTests : SenderReceiverClientTestBase
...
             try
             {
-                await this.ReceiveShouldReturnNoLaterThanServerWaitTimeTestCase(queueClient.InnerSender, queueClient.InnerReceiver, messageCount);
+                await this.ReceiveShouldReturnNoLaterThanServerWaitTimeTestCase(queueClient.InnerReceiver);
```
nemakam: Based on previous comment, remove the 'InnerReceiver' .
Same for all other occurrences

test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverClientTestBase.cs
```diff
@@ -124,7 +124,7 @@ public abstract class SenderReceiverClientTestBase
             TestUtility.Log("Sleeping 10 seconds...");
             await Task.Delay(TimeSpan.FromSeconds(10));
 
-            DateTime lockedUntilUtcTime = await messageReceiver.RenewLockAsync(receivedMessages.First().LockToken);
+            var lockedUntilUtcTime = await messageReceiver.RenewLockAsync(receivedMessages.First().LockToken);
```
nemakam: I don't think `var` should ALWAYS be used. For me, I prefer `var` wherever the type is understood and obvious when you don't have prior knowledge of the whole system. `var size = 1` is obvious. But in this case unless you open this in a IDE, you cant be sure whether the object returned is DateTime / DateTimeOffset or any other custom implementation.

SeanFeldman: @nemakam Agree. With `lockedUntilUtcTime` I would argue you're dealing with a date time that cares for UTC and not an offset. If you need to emphasize the offset, you'll do so with a variable name. For example: `localTimeWithOffset` or `localTimeWithTimezone` would indicate that you're looking at a time with offset component. 

Now, for the argument
> But in this case unless you open this in a IDE, you cant be sure whether the object returned is DateTime / DateTimeOffset or any other custom implementation.

The codebase is no longer a few files. To actually read it and understand, a review of a diff on GH is not enough. It either requires to pull the code down and see it all, or requires to be familiar with the code. I'd say if you pull the code down, you will be using an IDE.

We need to understand how vehicles operate to drive those, but do we really need to know to the low level how things work exactly? You know there's an engine, not Engive v6 3.5L blah blah blah. I see `var` as engine. Specifications (if needed) can be a part of the variable name.

Hopefully I have managed to convey my POV 😃

SeanFeldman:
@nemakam I've rebased my branch on top of `upstream/exposure` so that we don't work on two different PRs? Changed GH to target `upstream/dev` with this PR. I hope that's not an issue.

test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverClientTestBase.cs
nemakam: @SeanFeldman, this particular instance might fall in the grey area because of the naming convention of the variable. And I guess this is one of those never ending coding practices debate. Personally, I'd prefer that from next set of PRs, we use `var` with discretion and not everytime ReSharper suggests it :) 

Regarding
> The codebase is no longer a few files. To actually read it and understand, a review of a diff on GH is not enough. It either requires to pull the code down and see it all, or requires to be familiar with the code. I'd say if you pull the code down, you will be using an IDE.

I neither pull down the code and do the review in IDE, nor do I claim to be an expect with the entire code...

And regarding
> I see `var` as engine

I'm just looking out for people who see it as `object` and not engine.

SeanFeldman: Fair point. Agree to take it case by case.
R# can be harmful sometimes. That's why it's just a tool, not a developer replacement 😆

nemakam:
@SeanFeldman, Sure. Not an issue. I wonder why take the trouble of this when its going to get squashed anyway.

SeanFeldman:
An "opsie" moment? I realized that it was a bit too early after I've pushed...
Since we'll target with this PR `dev` anyways, no harm is done I thought. Sorry for that 😊

SeanFeldman:
Fixed the build. 💚

jtaubensee:
Hi @SeanFeldman - For simplicity's sake, lets close this PR, and reopen once #82 is merged. Does that work for you? I think it will make the history a little bit more clear.

SeanFeldman:
👍 or just cherry pick those changes 😉

PR #86, opened 2017-02-04
Updated roadmap

jtaubensee:
Updated roadmap and also adding dev appveyor build status link


jtaubensee:
Haha that is a lot to ask... This roadmap will probably go away once we reach GA, and is really only here to let people know what features are/are not available. We will adopt a model like [Event Hubs](https://github.com/Azure/azure-event-hubs-dotnet/releases) once we have a NuGet package.

Issue #34, opened 2016-12-14
OnMessage Handler


jtaubensee:
@vinaysurya to provide more details

vinaysurya:
API Surface

public void OnMessageAsync(Func<BrokeredMessage, Task> callback);
public void OnMessageAsync(Func<BrokeredMessage, Task> callback, OnMessageOptions onMessageOptions);

OnMessageOptions link:
	https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.onmessageoptions
	
Defaults:
    MaxConcurrentCalls is 1
	AutoRenewTimeout is set to 5 mins
	AutoComplete is set to true
	
Other User facing Contracts:
	OnMessageOptions.ExceptionReceived Event handler should be called when an operation in the Pump fails (say while receiving message/renewing lock on message/completing the message if AutoComplete is enabled/Abandon fails/ User Callback fails)
	
	The ExceptionReceived is a FYI only contract that is primarily used for debugging purposes to identify what kind of failures are hapenning on the Pump. The Pump should handle all the renew/abandon/retry operations internally during failures.
	
Description:
	The OnMessageAsync handler will take a User supplied callback and create a new MessageReceiverPump on its invocation. The Pump, as long as the receiver(associated with the QueueClient) is Open will keep trying to receive the message and for each message received it will in a seperate thread call for the Dispatch Task of that message. The Dispatch will inturn Schedule a RenewLock task for the message (on a different thread with a cancellation token) as specified by the OnMessageOptions and then call the User specified callback. Once the User callback is completed the Dispatch will cancel the renewloop task using the cancellation token.  If there is an exception with executing User Callback etc, the ExceptionReceived event handler has to be called. Additionally, if the QueueClient is PeekLock, if the User call back has succeeded and AutoComplete is turned on, the message needs to be completed, else the message will need to be abandoned.

vinaysurya:
@SeanFeldman, @jtaubensee  I've added more details for the OnMessageHandler. Please let me know if you need any more details

SeanFeldman:
@vinaysurya this is great.

1. When `OnMessageOptions.ExceptionReceived` event handler is invoked, it doesn't provide information what  namespace it's using. Having that information in the new client implementation would help for custom fail-over implementation.

[...]

PR #92, opened 2017-02-25
On message feature

vinaysurya:
## Description


src/Microsoft.Azure.ServiceBus/OnMessageOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    using System;
+    using Primitives;
+
+    public sealed class OnMessageOptions
+    {
+        int maxConcurrentCalls;
+        TimeSpan autoRenewTimeout;
+
+        /// <summary>Initializes a new instance of the <see cref="OnMessageOptions" /> class.</summary>
```
SeanFeldman: It would be good for the code to document the defaults.
Otherwise, it's hoping that msdn / doco will get updated, which is always a PITA.

SeanFeldman:
@vinaysurya would be nice to mark in the general description what issue(s) a PR is addressing.
In this case it's #34

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -0,0 +1,266 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.ServiceBus.Primitives;
+
+    public sealed class MessageReceivePump
```
vinaysurya: did not need to be public, removed

src/Microsoft.Azure.ServiceBus/Primitives/TaskExtensionHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Primitives
+{
+    using System;
+    using System.Threading.Tasks;
+
+    public class TaskExtensionHelper
```
vinaysurya: does not need to be public, removed

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -0,0 +1,266 @@
...
+                    MessagingEventSource.Log.MessageReceiverPumpInitialMessageReceived(this.messageReceiver.ClientId, initialMessage);
+                    break;
+                }
+                catch (Exception exception)
+                {
+                    retryCount++;
+                    MessagingEventSource.Log.MessageReceiverPumpInitialMessageReceiveException(this.messageReceiver.ClientId, retryCount, exception);
+
+                    if (retryCount == MaxInitialReceiveRetryCount ||
+                        !this.ShouldRetry(exception))
```
vinaysurya: I dont think its a significant improvment over, ShouldRetry(exception)

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -0,0 +1,266 @@
...
+                    {
+                        throw;
+                    }
+
+                    TimeSpan backOffTime = this.GetBackOffTime(exception);
+                    await Task.Delay(backOffTime, this.pumpCancellationToken).ConfigureAwait(false);
+                }
+            }
+
+            TaskExtensionHelper.Schedule(() => this.MessagePumpTask(initialMessage));
```
vinaysurya: This method tries to execute a method in a FireAndForget kind of mechanism, as in I want to execute this but do not wait for it to finish. Doing this just as is, without a corresponding await, will result in a CS4014, "because await is not called.. " warning.. Hence an extension to schedule this method to run in the context of another task.

src/Microsoft.Azure.ServiceBus/Primitives/TaskExtensionHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public static void Schedule(Func<Task> func)
+        {
+            Task.Run(async () =>
+            {
+                try
+                {
+                    await func();
```
vinaysurya: Thanks!

src/Microsoft.Azure.ServiceBus/Primitives/TaskExtensionHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+                }
+                catch (Exception)
+                {
```
vinaysurya: If you look at how the inner tasks that are passed to this function are written, all of them are self contained, in the sense that they try catch everything that happen in the function internally.  So I did not see a real need for logging here again. But in any case, it might be a good idea to log, I'll add a todo to log any unexpected exception here.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -0,0 +1,266 @@
...
+                    {
+                        this.onMessageOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, "RenewLock"));
+                    }
+                    if (!this.ShouldRetry(exception))
+                    {
+                        break;
+                    }
+
+                    TimeSpan backoffTimeSpan = this.GetBackOffTime(exception);
+                    await Task.Delay(backoffTimeSpan).ConfigureAwait(false);
```
vinaysurya: done

src/Microsoft.Azure.ServiceBus/OnMessageOptions.cs
```diff
@@ -0,0 +1,84 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus
+{
+    using System;
+    using Primitives;
+
+    public sealed class OnMessageOptions
```
vinaysurya: done

src/Microsoft.Azure.ServiceBus/OnMessageOptions.cs
SeanFeldman: @vinaysurya what about this one?

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
SeanFeldman: You can still do that w/o loosing the connection to the original code by defining an extension method:

```c#
static class TaskExtentions
{
    public static void Ignore(this Task task)
    {
        task.ConfigureAwait(false);
    }
}
```
and your code becomes:

``` c#
Task.Run(async () =>
{
  try
  {
    await this.MessagePumpTask(initialMessage).ConfigureAwait(false);
  }
  catch (Exception)
  {
    // TODO: Log any unexpected exception here.
  }
}).Ignore();
```

jtaubensee: Also, we can definitely turn off the CS4014, since there are times where an async void is useful...

vinaysurya: Even with the way you suggest, it is still making an additional call right? I don't see how this saves an additional call.. Or are you talking about readability here? Besides, with the way Schedule is written wouldn't the JIT compiler inline it anyway, so it wouldn't result in an extra call everytime

vinaysurya: @jtaubensee, yes we could do that or I could supress the warning on the method too, but since there was an easy workaround here, I did not want to go that route. I still think it might be useful to have that warning and make people forcefully override it on a case by case basis if needed

SeanFeldman: Readability more than anything else. Up to you how you decide to proceed 😄

Issue #87, opened 2017-02-10
Simple vs. advanced API set


jtaubensee:
Created a new issue from #57.

Since this is a new client, our team has taken some time to reevaluate the current usage of our past library, and we want to make this new one as simple as possible. In order to do so, we are planning to use clients (i.e. Queue, Topic, Subscription) for general messaging scenarios, while using message senders/receivers for more advanced scenarios. As such we are proposing the following:

* Keep clients in the root namespace, and move sender/receivers to a separate namespace. We would like to keep the root namespace as simple as possible, allowing new customers to get started quickly without the need to learn too many concepts.
* Limit clients to a message pump model (i.e. `OnMessage`), and reserve direct receive calls for the `MessageReceiver`.
* Remove static create methods from the clients, and instantiate them through constructors.
* Some sort of connection pooling for clients, and allow customers to implement their own pooling algorithms for senders/receivers (this would come much later though).

After some initial investigation, the QueueClient interface can look something like this:
```csharp
    public interface IQueueClient
    {
        public IQueueClient(string connectionString, ReceiveMode mode);   
        public string QueueName;
        public ReceiveMode Mode;
        public int PrefetchCount;    
        public Task CloseAsync();
        public Task SendAsync(BrokeredMessage message);
        public Task SendAsync(IList<BrokeredMessage> messages);
        public Task CompleteAsync(Guid lockToken);
        public Task AbandonAsync(Guid lockToken);
        public Task DeferAsync(Guid lockToken);
        public Task DeadLetterAsync(Guid lockToken);
        public Task<long> ScheduleMessageAsync(Message message, DateTimeOffset scheduledEnqueueTimeUtc);
        public Task CancelScheduledMessageAsync(long sequenceNumber);
        public void RegisterMessageHandler(OnMessageOptions options, Func<BrokeredMessage, Task>);
        public void RegisterSessionHandler(OnSessionOptions options, Func<Session, Task>);
    }
```
Other functionality that exisits in the .NET API today, that is not in the QueueClient above would then live in either `MessageSender` or `MessageReceiver`.

SeanFeldman:
> Keep clients in the root namespace, and move sender/receivers to a separate namespace. We would like to keep the root namespace as simple as possible, allowing new customers to get started quickly without the need to learn too many concepts.

> Remove static create methods from the clients, and instantiate them through constructors.

[...]

Issue #57, opened 2017-01-03
Can't use MessageSender & MessageReceiver


SeanFeldman:
## Expected Behavior
Use `MessageSender` and `MessageReceiver` to send/receive messages. 

## Actual Behavior
Classes are abstract and cannot be used.

Version: [current master](https://github.com/Azure/azure-service-bus-dotnet/blob/fc95ccf00125fd7046b121aeb8ac5b025c1454c5/src/Microsoft.Azure.ServiceBus/MessageReceiver.cs#L12)
Commit: https://github.com/Azure/azure-service-bus-dotnet/commit/a34fbdb62eee0387ab9b08844c089a8ce5e32e26

binzywu:
Hi @SeanFeldman, could you please elaborate a little more about the scenarios that you will need this?

SeanFeldman:
Absolutely @binzywu 
When a process is receiving messages, it handles messages from queues and subscriptions. The process doesn't differentiate the entity type messages come from as messages contain information required for proper processing. Same is applied for sending, messages are sent out using a path, w/o differentiating if destination is a topic or a queue.

With both `MessageSender` and `MessageReceiver` abstract it is not possible with the new client.

binzywu:
@SeanFeldman, @nemakam will be working on this.

SeanFeldman:
👍 @binzywu 
I'll be following the PR the moment it shows up.

nemakam:
You can track the changes in PR #82

jtaubensee:
Closing in favor of #87

nemakam:
Reopening the issue since MessageSender and MessageReceiver are made internal in #87 .

jtaubensee:
Also, a reminder to rename the `Peek()` operation to `Browse()` in order to avoid confusion between `PeekLock` mode.

SeanFeldman:
:-1: for `Browse`.
`Peek` is a well established term for the operation and means exactly what it does. Not to mention that other messaging and queueing technologies use the same nomenclature.

jtaubensee:
Closing based on #146

PR #95, opened 2017-03-02
Simple vs. advanced API set 

nemakam:
## Description
Simple vs. advanced API set (Refer Issue #87)
1. Renamed BrokeredMessage to Message
2. QueueClient is no longer abstract. You can now write 'new QueueClient()' which returns the default AMQPQueueClient. This basically means we can no longer have AMQPQueueClient as an inherited child class of QueueClient. Amqp related things have now been moved inside the object as QueueClient.InnerClient now. 
3. Moved internal entities into ServiceBus.Core namespace.
4. Externally, lockTokens are strings instead of Guids
5. OnMessage has been rebranded to RegisterMessageHandler
6. MessageSender and MessageReceivers have been made internal for now.
7. Sessions are currently not available on the Clients and will be replaced with 'OnSession' equivalent soon.


src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Core
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface IReceiverClient : IClientEntity
```
SeanFeldman: Public API should have xmldoc

src/Microsoft.Azure.ServiceBus/Core/ISenderClient.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Core
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    public interface ISenderClient : IClientEntity
```
SeanFeldman: Public API should have xmldoc

src/Microsoft.Azure.ServiceBus/IMessageSession.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus
+{
+    using System;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Core;
+
+    internal interface IMessageSession : IMessageReceiver
```
SeanFeldman: Public API should have xmldoc

src/Microsoft.Azure.ServiceBus/Primitives/IClientEntity.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus
+{
+    using System.Threading.Tasks;
+
+    public interface IClientEntity
```
SeanFeldman: Public API should have xmldoc

SeanFeldman:
> MessageSender and MessageReceivers have been made internal for now.

@nemakam does that mean it's in the scope of this PR to make it public again? If not, then should have an issue to track that.

samples/SendSample/Program.cs
```diff
@@ -46,7 +39,7 @@ public static void Main(string[] args)
                 try
                 {
                     // Create a new brokered message to send to the queue
-                    var message = new BrokeredMessage($"Message {i}");
+                    var message = new Message($"Message {i}");
```
nemakam: BrokeredMessage is an internal concept which doesn't need to be used publicly. Hence the reason we are changing it. I agree there might be a small number of cases when the word Message might be used already, but that might not be the majority scenario. And even in this case, the issue can be resolved using namespace prefix. Since this is a new API, we plan to have the nomenclature crisp and avoid all the internal details

src/Microsoft.Azure.ServiceBus/Core/IInnerSender.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Azure.ServiceBus.Core
+{
+    using System.Threading.Tasks;
+
+    internal interface IInnerSender
+    {
+        MessageSender InnerSender { get; }
+
+        MessageSender CreateMessageSender();
```
nemakam: True. It was kept for extensibility purposes in the first place. Maybe someday if you ever need to expand it to beyond AMQP. But given the future plans, I dont' see it happening anytime soon. So removing it.

src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
nemakam: We are yet to completely finalize on all the APIs for clients as well as Receivers and Senders. Will take up documentation separately before releasing the package.

nemakam:
@SeanFeldman , Making the two classes public is outside the scope of this PR. I have reopened Issue #57 .

src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
SeanFeldman: Sometimes writing the doco along with the code raises good questions. Not going to insist on this, but will remind from time to time 😉

SeanFeldman:
@nemakam thank you for re-opening #57

Issue #98, opened 2017-03-09
Change Message.Body type to byte[]


jtaubensee:
## Actual Behavior
* In the [.NET framework client](https://www.nuget.org/packages/WindowsAzure.ServiceBus/) a `BrokeredMessage` can be a serialize-able object, or a stream.

## Expected Behavior
* Change `Message` to only support `byte[]`
* We are proposing that the `Message` class be closely related to the Event Hubs [`EventData`](https://github.com/Azure/azure-event-hubs-dotnet/blob/dev/src/Microsoft.Azure.EventHubs/EventData.cs) class

## Why
* The use of streams makes extensibility much more difficult. The `Message` class is already a heavier object than it needs to be, and a majority of the logic in the class is around stream handling. For instance, if we want to make a handle to manipulate the `Message.Body` when sending or receiving a message, we need to be able to handle multiple streams, and "resetting" the body. This proposal would allow you to simply call something like: `message.Body = new byte[]();`
* Passing streams to Service Bus can result in a blocking send. For example, if you send a message with a stream and call send, the entire stream needs to be read (from where ever it is) into memory before the actual send occurs. This can cause users to think that there is an issue with Service Bus performance, when that is not actually the case.

### Edit:
Pointed out by @seanfeldman, this also means that we would remove the `DataContractSerializer` pieces as well.

SimonCropp:
> the entire stream needs to be read (from where ever it is) into memory before the actual send occurs

why is this the case? cant the data be streamed all the way to the remote receiver? meaning the full byte array would never need to exist in memory

jtaubensee:
@SimonCropp - Yes, that is correct. [Here](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs#L371) we actually call [StreamToBytes](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs#L542) which will convert any stream to a `byte[]`.

SimonCropp:
cant the data be streamed all the way to the remote receiver? meaning the full byte array would never need to exist in memory

jtaubensee:
AMQP doesn't actually accept `Stream` as a data type, so our implementation of it in the past was just a "nice" feature for users. So either way we are holding the message in memory.

SimonCropp:
[...]

PR #108, opened 2017-03-22
(Brokered)Message refactoring

jtaubensee:
## Description
Addresses #98
BrokeredMessage refactoring:
* Changed BrokeredMessage -> Message and simplified the object.
  * Removed stream/serialization functionality
  * Body is an ArraySegment<byte>
  * Moved "Service Bus properties" to SystemProperties
  * Moved "user properties" to UserProperties
  * Removed `MessageIdGenerator`. To be added as a plugin (@seanfeldman)
* Updated dependencies


jtaubensee:
## Items still needing work:
* Message.Size: Need to provide support for body / size calculation
* Additional unit tests for Message and AMQPConverter

jtaubensee:
passing tests

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,200 @@ namespace Microsoft.Azure.ServiceBus
...
-        /// <summary>Gets the number of deliveries.</summary>
-        /// <value>The number of deliveries.</value>
-        /// <exception cref="System.ObjectDisposedException">Thrown if the message is in disposed state.</exception>
-        /// <exception cref="System.InvalidOperationException">Thrown if the message has not been delivered by ServiceBus.</exception>
-        public int DeliveryCount
-        {
-            get
-            {
-                this.ThrowIfDisposed();
-                this.ThrowIfNotReceived();
```
jtaubensee: Should this be removed? i.e. Should the default behavior be to throw when trying to access a property that would only be set when a message is received?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,200 @@ namespace Microsoft.Azure.ServiceBus
...
+            return clone;
         }
 
         /// <summary> Validate message identifier. </summary>
         /// <exception cref="ArgumentException">
         /// Thrown when messageId is null, or empty or greater than the maximum message length.
         /// </exception>
         /// <param name="messageId"> Identifier for the message. </param>
-        static void ValidateMessageId(string messageId)
+        private void ValidateMessageId(string messageId)
```
SeanFeldman: Any specific reason why this shouldn't be static?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -1150,7 +214,7 @@ static void ValidateMessageId(string messageId)
         /// Thrown when sessionId is greater than the maximum session ID length.
         /// </exception>
         /// <param name="sessionId"> Identifier for the session. </param>
-        static void ValidateSessionId(string sessionId)
+        private void ValidateSessionId(string sessionIdPropertyName, string sessionId)
```
SeanFeldman: Why not static?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -1159,7 +223,7 @@ static void ValidateSessionId(string sessionId)
             }
         }
 
-        static void ValidatePartitionKey(string partitionKeyPropertyName, string partitionKey)
+        private void ValidatePartitionKey(string partitionKeyPropertyName, string partitionKey)
```
SeanFeldman: Why not static?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -1168,234 +232,26 @@ static void ValidatePartitionKey(string partitionKeyPropertyName, string partiti
...
-            {
-                // TODO: throw Fx.Exception.ObjectDisposed("Message has been disposed.");
-                throw new ObjectDisposedException("Message has been disposed.");
-            }
-        }
-
-        /// <summary> Throw if not locked. </summary>
-        /// <exception cref="FxTrace.Exception"> Thrown when as error. </exception>
-        void ThrowIfNotLocked()
+        public sealed class SystemPropertiesCollection
```
SeanFeldman: It says collection, but it's more like `SystemProperties`.

SeanFeldman: Deserves a file of its own? 😃

test/Microsoft.Azure.ServiceBus.UnitTests/AmqpConverterTests.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    using System;
+    using System.Text;
+    using Microsoft.Azure.ServiceBus.Amqp;
+    using Xunit;
+    using BrokeredMessage = Microsoft.Azure.ServiceBus.Message;
+
+    public class AmqpConverterTests
+    {
+        public class WhenConvertingBrokeredMessagesToAmqpMessages
```
SeanFeldman: `WhenConvertingBrokeredMessagesToAmqpMessages` vs `When_converting_brokered_messages_to_amqp_messages` I'll take 2nd option any time as it's easier to read. If this was a German word, that would be a different story 😉

samples/ReceiveSample/Program.cs
```diff
@@ -40,7 +41,7 @@ private static void ReceiveMessages()
                     async (message, token) =>
                     {
                         // Process the message
-                        Console.WriteLine($"Received message: SequenceNumber:{message.SequenceNumber} Body:{message.GetBody<string>()}");
+                        Console.WriteLine($"Received message: SequenceNumber:{message.SystemProperties.SequenceNumber} Body:{Encoding.UTF8.GetString(message.Body.Array)}");
```
SeanFeldman: `message.Body.Array` feels like friction.

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,200 @@ namespace Microsoft.Azure.ServiceBus
...
+            this.SystemProperties = new SystemPropertiesCollection();
+            this.UserProperties = new Dictionary<string, object>();
         }
 
-        /// <summary> Gets or sets the the Publisher. </summary>
-        /// <value> Identifies the Publisher Sending the Message. </value>
-        /// <exception cref="ObjectDisposedException">Thrown if message is in disposed state.</exception>
-        /// <exception cref="InvalidOperationException">Thrown if <seealso cref="PartitionKey"/> or <seealso cref="SessionId"/> are set to different values.</exception>
-        internal string Publisher
+        public ArraySegment<byte> Body { get; set; }
```
SeanFeldman: Any specific reason why `Body` is of type `ArraySegment<byte>` and not `byte[]`? Manipulations would be done on the entire body, not subsets of bytes, so what is the benefit?

SeanFeldman:
Are we going back to the `BrokeredMessage`?

src/Microsoft.Azure.ServiceBus/Message.cs
jtaubensee: The idea is that if you need to split up a larger message into multiple smaller messages, this would be simpler. Followed guidance [here](https://msdn.microsoft.com/library/1hsbd92d(v=vs.110).aspx#Remarks). I think that ArraySegment is probably OK, but since a lot of customers deal with strings, we should provide and overload for that, and an easy way to access the body as a string.

src/Microsoft.Azure.ServiceBus/Message.cs
jtaubensee: Does it need to be static? Its a private method...

jtaubensee:
@SeanFeldman - I don't think we are going back to BrokeredMessage, I just used the alias to make it clear what the AMQP converter was doing. I can just as easily use an alias like "SBMessage" in that class too.

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,200 @@ namespace Microsoft.Azure.ServiceBus
...
-    using System.IO;
-    using System.Runtime.Serialization;
-    using System.Threading;
-    using Core;
-    using Primitives;
+    using System.Text;
 
-    /// <summary>Represents the unit of communication between ServiceBus client and Service.</summary>
-    public sealed class Message : IDisposable
+    public class Message
```
jtaubensee: per @vinaysurya this should still implement IDisposible. How about being sealed?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,200 @@ namespace Microsoft.Azure.ServiceBus
...
-                {
-                    clonedStream = (Stream)cloneable.Clone();
-                }
-                else if (canThrowException)
-                {
-                    // TODO: throw Fx.Exception.AsError(new InvalidOperationException(SRClient.BrokeredMessageStreamNotCloneable(originalStream.GetType().FullName)));
-                    throw new InvalidOperationException("BrokeredMessageStreamNotCloneable");
-                }
+                var clonedBody = new byte[this.Body.Count];
+                Array.Copy(this.Body.Array, clonedBody, this.Body.Count);
```
nemakam: This doesn't copy as intended.
ArraySegment has an offset and a count. The offset need not be 0.

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: @SeanFeldman , do you see any particular concern with keeping the class definition in the same file? This class is never used outside of Message context and should not be.

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: Can you please add a test which initializes a message with a true arraySegment (part of byte[] which neither starts or ends at the start or end of the byte[]). You need to test for send, receive and a clone for that.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: So that makes sense in the context of the application that is handling the body as a source of multiple messages. For the general purpose, you could return `byte[]`. 

For the case where you need to operate on sections of the body it would be `new ArraySegment<byte>(msg.Body)`. Why to impose `ArraySegment<byte>` on all use cases?

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: For performance. A single static method as it doesn't need to access any non-static members.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: It's much easier to navigate to the class when the filename represent the class and not a container class. Not as critical as the name comment.

nemakam: @SeanFeldman, true for the navigation. But with advent of IDEs, I think it shouldn't be that big of a concern. In contrast, if you do not split the file, then you have a logical closure on where the class is being used. I personally would prefer having it in same class in this scenario when it strictly doesn't make sense to be used outside the scope of a Message

SeanFeldman: For classes that are not public, perhaps. For public stuff I'd push back. The closure can be provided by the the IDEs and tooling such as VS2017 and R#. GH PRs are often reviewed in the browser. Navigation by file type is allowed on all branches. Search is only on the active branch. So navigating to `SystemPropertiesCollection` would not be possible. But again, this is not critical, just a suggestion. Name is more important.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: @nemakam you mean `Array.Copy(src.Array, src.Offset, dst, 0, src.Count);`?
I'm still not convinced this is needed... if you send a giant payload and chunk it, it's the application responsibility, not the brokered message. Brokered message should just accept a `byte[]`.

nemakam: I'd also personally prefer using byte[]. But this comment was keeping in mind if ArraySegment is chosen. And yes, your overload of Array.Copy looks right.

SeanFeldman: It feels like `ArraySegment<byte>` is more of an application concern, not the brokered message. Brokered message should deliver the bytes and the app can choose how to process/operate on those.

src/Microsoft.Azure.ServiceBus/Message.cs
jtaubensee: So if we keep it in the same file and rename the class to just `SystemProperties` there is a name clash between that and the property on `Message`. I am of the opinion to keep the class in the file, and therefore the name would stay the same. I personally like the clarity of 'this class is only ever used in here'

jtaubensee:
Latest set of changes:
- Changed `Message.Body` from `ArraySegment<byte>` to `byte[]`
- Added `Message.BodyAsString` method which provides the byte[] body as a UTF8 encoded string
- Moved `Message.LockToken` to `Message.SystemProperties.LockToken`
- Changed `BrokeredMessage` alias inside of `AmqpConverter` to `SBMessage` (letting BrokeredMessage rest peacefully :smile:).

src/Microsoft.Azure.ServiceBus/Message.cs
jtaubensee: https://msdn.microsoft.com/library/79b3xss3(v=vs.110).aspx tldr;
>A call to a static method generates a call instruction in Microsoft intermediate language (MSIL), whereas a call to an instance method generates a callvirt instruction, which also checks for a null object references. However, **most of the time** the performance difference between the two is **not significant**.

SeanFeldman: While you might not get a lot of perf, it is on the hot path (executed per each message going out). If there's no state, really no reason to keep it non-static.

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -6,1136 +6,202 @@ namespace Microsoft.Azure.ServiceBus
...
+        }
 
-                this.ThrowIfDisposed();
-                this.initializedMembers |= MessageMembers.PartitionId;
-                this.partitionId = value;
-            }
+        public Message(byte[] array)
+        {
+            this.Body = array;
+            this.SystemProperties = new SystemPropertiesCollection();
```
SeanFeldman: Collection implies a set of item of the same time, where here it's just a group of properties. Should it be just a `SystemProperties` type?

jtaubensee: So if we keep it in the same file and rename the class to just SystemProperties there is a name clash between that and the property on Message. I am of the opinion to keep the class in the file, and therefore the name would stay the same. I personally like the clarity of 'this class is only ever used in here'

SeanFeldman: @jtaubensee not following... there's no clash.
If you have 2 files (`Message.cs` and `SystemProperties.cs`) or not, it is valid for `Message` class to contain a property of `SystemProperties` type and be named `SystemProperties`.

```c#
    public class SystemProperties
    {
        public int SequenceNumber { get; set; }
    }
```

```c#
    public class Message
    {
        public SystemProperties SystemProperties { get; set; }
    }
```

jtaubensee: This happens when `SystemProperties` is in the same class:

2017-03-23_10h58_51

jtaubensee: Based on my earlier comment, I'd prefer to leave it in the same file. I think it is more clear that the `SystemProperties` class is used specifically in `Message`.

SeanFeldman: I'm not sure why you get that error... this is what I see in a quick spike:
image

nemakam: The difference would be Innerclass vs "two classes in the same file"...

SeanFeldman: Why would `SystemProperties` be an inner class of `Message` if it's public I do not follow...anyhow let it be suffixed with `Collection` for fun 😃

src/Microsoft.Azure.ServiceBus/Message.cs
jtaubensee: @SeanFeldman / @nemakam / @vinaysurya - Any thoughts here?

SeanFeldman: Makes sense to keep it. If message was never received (ie locally created message), `DeliveryCount` is not valid. Alternative (if you don't want to throw) would be to return an invalid value (-1?). I'm not sure what would be better. 

Side note: I've added `.IsReceived` on the message for (PR #78) to avoid exceptions when dealing with messages w/o the knowing if it was received or created locally.

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: @SeanFeldman, Just found this online: https://coderanch.com/t/579241/java/public-nested-classes-good-practice. 
I'm sure there might be better articles than this online...

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: I'm also in favor of throwing here. Different data types will have different 'invalid value', and might make it tricky. Throwing exception in conjunction with 'IsReceived' property sounds like a good approach

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: @nemakam not sure. All depends on the context.

jtaubensee:
Created #109 to track Message.Size functionality

SeanFeldman:
@jtaubensee should this be tracked separately as well?

> https://github.com/MessageHandler/MessageHandler.Fabric/pull/2

niemyjski:
Do we even need (Message.BodyAsString). I'd rather keep this stuff really clean, if you want it as a string you could easily do that yourself.

nemakam:
@niemyjski , I would still prefer having the string overload for set-get of Body as sending string messages is one of the most common use-cases. I'm sure its a do-able thing for the customer himself, but having the overload makes it easier and makes the code look better, as compared to doing `Encoding.UTF8.GetString()` everytime.

niemyjski:
Is it lazy loaded? or do we have to take that allocation hit every time even if we don't need or want it?

nemakam:
@niemyjski. it is being implemented as `public string BodyAsString => Encoding.UTF8.GetString(this.Body);`. This should be lazy loaded.

niemyjski:
@nemakam shouldn't this be stored in a private member the first time you access it? If you call that multiple times that's a lot of work each time you call BodyAsString

SeanFeldman:
@niemyjski why would you application call this property multiple times?

Assuming your code is using `BodyAsString` multiple times, I'd counter argue with the following: if you create a message locally, and modify it, having a cached body would fire back. To address that, you'd need to introduce a staleness check, which makes the whole thing way more complicated than it should be.

niemyjski:
Your message is immutable and shouldn't ever change. I've seen sample code as well as code in prod where bodyAsString/similar property is called multiple times without realizing what is happening under the covers.

SeanFeldman:
Personally, I think this `BodyAsAnything` isn't necessary since `byte[]` is all I care for. There's no `BodyAsStream` or other variant, so what's the point? KISS principle 😃

niemyjski:
@SeanFeldman I agree!

jtaubensee:
@niemyjski / @SeanFeldman - Sorry for the delay, I was out of the office last week. To clarify, the body is definitely mutable with this client, however, I dont have any problem removing the string over load functionality for Message.Body, as we can gauge user feedback and easily add it at anytime.

SeanFeldman:
@jtaubensee agreed. If it's a needed feature, people will ask/vote for it.

src/Microsoft.Azure.ServiceBus/Message.cs
ApolloCreek: … if so, then embed the class in its only using class. Otherwise move it to an own file!

PR #110, opened 2017-03-30
Retrypolicy

nemakam:
## Description
Implemented an abstract RetryPolicy class with two implementations: NoRetry and RetryExponential
(Issue #7)


test/Microsoft.Azure.ServiceBus.UnitTests/RetryTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            var retry = RetryPolicy.Default;
+            Assert.True(retry is RetryExponential);
+        }
+
+        [Fact]
+        void RetryExponentialShouldRetryTest()
+        {
+            // Tuple<ExceptionType, CurrentRetryCount, ShouldRetry>
+            Tuple<Exception, int, bool>[] listOfExceptions =
```
SeanFeldman: Could you please switch to `[Theory][MemberData]` approach similar to what is already done [here](https://github.com/Azure/azure-service-bus-dotnet/blob/f2a4b1bfa7f4073f83210ad2e66b06366e147981/test/Microsoft.Azure.ServiceBus.UnitTests/QueueClientTests.cs#L19)? It will make the test easier to follow and see what fails individually, rather than the whole bunch.

test/Microsoft.Azure.ServiceBus.UnitTests/RetryTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            var policy = RetryPolicy.Default;
+            Stopwatch watch = Stopwatch.StartNew();
+            try
+            {
+                await policy.RunOperation(() => { throw new ServiceBusException(true, string.Empty); }, TimeSpan.FromSeconds(3));
+            }
+            catch (Exception)
+            {
+                // Expected
```
SeanFeldman: If expected, shouldn't it be verified?

test/Microsoft.Azure.ServiceBus.UnitTests/RetryTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        async void RunOperationShouldWaitForAllOperationsToSucceed()
+        {
+            var policy = RetryPolicy.Default;
+            Stopwatch watch = Stopwatch.StartNew();
+            try
+            {
+                await policy.RunOperation(
+                    async () =>
+                    {
+                        for (int i = 0; i < 5; i++)
```
SeanFeldman: Why the loop is needed? What if was a single delay of 10sec? 🤔

test/Microsoft.Azure.ServiceBus.UnitTests/TestConstants.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.Azure.ServiceBus.UnitTests
 {
-    static class Constants
+    static class TestConstants
```
SeanFeldman: 👍

test/Microsoft.Azure.ServiceBus.UnitTests/RetryTests.cs
nemakam: It was testing pitfalls of async-programming. Refer https://blogs.msdn.microsoft.com/pfxteam/2012/02/08/potential-pitfalls-to-avoid-when-passing-around-async-lambdas/#comments

SeanFeldman: @nemakam the link you're posted talks about the danger of passing an asyncy lambda into an `Action` will become a void call. The solution outlined within the post clearly states that the proper way is to use a `Func<Task>` instead of `Action`. And that is already done in your code and creating a loop doesn't change anything. Let me trace it here:

You have an async lambda passed into [`RetryPolicy.RunOperation()`](https://github.com/Azure/azure-service-bus-dotnet/pull/110/files/e93fef47f9d7fd1bf5ccff9d55d20d392dc09347#diff-c4777b8b45d84d6fbac7b96430200333R36) which _already_ takes in a `Func<Task>`. Internally within the method, you are [awaiting for the `operation`](https://github.com/Azure/azure-service-bus-dotnet/pull/110/files/e93fef47f9d7fd1bf5ccff9d55d20d392dc09347#diff-c4777b8b45d84d6fbac7b96430200333R59). 

So having multiple delays is not going to change anything and you could just have a single delay. If you would like to validate that several tasks got completed, you should have task handlers validated that after ` await policy.RunOperation()` is called, all tasks are in `Completed` state 😺

nemakam: @SeanFeldman , True, i'm already taking care of it by using Fun<Task> instead of Action. But isn't that what automated test cases all about? Ensuring there is no regressions later? Ensuring that code isn't changed unknowingly next time without keeping this in mind?

And yes, ensuring tasks are in completed states is a valid approach. The current implementation also ensures that. Also ensures that multiple operations are not running in parallel which may happen with bad coding practices, which isn't too much to test, but an extra freebie is always good :)

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -64,11 +64,11 @@ public Task SendAsync(Message message)
             message.ScheduledEnqueueTimeUtc = scheduleEnqueueTimeUtc.UtcDateTime;
             MessageSender.ValidateMessage(message);
             MessagingEventSource.Log.ScheduleMessageStart(this.ClientId, scheduleEnqueueTimeUtc);
-            long result;
+            long result = 0;
```
nemakam: It actually gives out an error that I am returning an in-initialized variable

test/Microsoft.Azure.ServiceBus.UnitTests/RetryTests.cs
SeanFeldman: Time based testing is always tricky... if you could validate in a different way that there's no regression (`Func<Task>` to `Action`) I'd probably be a happier camper.

BTW, for the API changes, you could have a test that tests `RetryPolicy.RunOperation()`'s `operation` parameter to be of a specific type. And if it's not, throw a meaningful exception about regression. **That** would be very easy to understand 😉

PR #116, opened 2017-04-10
Fix of incomplete reading of the Body

ddobric:
Related to [Issue-17](https://github.com/Azure/azure-service-bus-dotnet/issues/114)

Please ignore or VS2017 related changes (sorry).

The only relavant change is in following [file](https://raw.githubusercontent.com/ddobric/azure-service-bus-dotnet/master/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs).

All other changes in this PR can be ignored.

`  else if (amqpMessage.BodyType == SectionFlag.Data && amqpMessage.DataBody != null)
            {
                var data = amqpMessage.DataBody.FirstOrDefault();
                if (data != null)
                {
                    // Chek cast ???
                    var seg = (ArraySegment<byte>)data.Value;
                    sbMessage = new SBMessage(seg.ToArray());
                }
                else
                {
                    // ??
                    sbMessage = null;                   
                }
            }`

This implementation is not complete. However it looks as we would miss a part of implementation here especially related to a way how SB .net desktop client serialize messages.

Issue #114, opened 2017-04-07
Interop support for other clients


ddobric:
Have upgraded to 0.0.2-preview and running in some issues.

1. QueueClient does not have a method GetBody<>() as described in samples.
2. After receiving of the message property 'Body' is always NULL.
3. There is no more ReceiveAsync() method

jtaubensee:
Hi @ddobric,

Please keep in mind that this is not a direct port of the older WindowsAzure.ServiceBus package. Since we are rewriting the library, we have taken the time to reorganize/redesign the api's and interfaces. For an example of this check out #87.

> QueueClient does not have a method GetBody<>() as described in samples

Which samples are you referring to?

> After receiving of the message property 'Body' is always NULL

Can you share some code?

> There is no more ReceiveAsync() method

Same as response in #13.

ddobric:
**Receiver is the original sample code from repo:**

`
  private static void ReceiveMessages()
        {
            try
            {
                // Register a OnMessage callback
                queueClient.RegisterMessageHandler(
                    async (message, token) =>
                    {
                        // Process the message
                        Console.WriteLine($"Received message: SequenceNumber:{message.SystemProperties.SequenceNumber} Body:{Encoding.UTF8.GetString(**message.Body**)}");

                        // Complete the message so that it is not received again.
                        // This can be done only if the queueClient is opened in ReceiveMode.PeekLock mode.
                        await queueClient.CompleteAsync(message.SystemProperties.LockToken);
                    },
                    new RegisterHandlerOptions() {MaxConcurrentCalls = 1, AutoComplete = false});
            }
            catch (Exception exception)
            {
                Console.WriteLine($"{DateTime.Now} > Exception: {exception.Message}");
            }
        }
`

**Sender is based on SB .NET (Desktop)  SDK:**

`  
private static void sendMessage()
        {
            var sender = QueueClient.CreateFromConnectionString(..);

            MyType msg = new MyType()
            {
              . . .              
            };

            sender.SendAsync(createMessage(msg)).Wait();
        }

        private static BrokeredMessage createMessage(object sendingObject)
        {
            string strMsg = JsonConvert.SerializeObject(sendingObject);
            BrokeredMessage msg = new BrokeredMessage(UTF8Encoding.UTF8.GetBytes(strMsg));
            msg.Properties.Add("XYZ", "THIS WORKS");
            return msg;
        }`

After receiving message,Body is NULL!!!

ddobric:
[...]

PR #118, opened 2017-04-10
Added ability to read AMQP DataBody in AMQPConverter

jtaubensee:
## Description
Replacing PR #116, and addresses #114.
Allows for receiving AMQP messages that have the AMQP Data Body set in addition to AMQP Value. This allows for the older Service Bus client to be used in conjunction with this library.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
@@ -177,9 +177,13 @@ namespace Microsoft.Azure.ServiceBus.Amqp
                 && amqpMessage.DataBody != null
                 && amqpMessage.DataBody.Count() > 0)
             {
-                var dataBody = amqpMessage.DataBody.FirstOrDefault().Value;
-                var dataBodyAsByteArray = (ArraySegment<byte>)dataBody;
-                sbMessage = new SBMessage(dataBodyAsByteArray.ToArray());
+                var dataSegments = new List<byte>();
+                foreach (Data data in amqpMessage.DataBody)
+                {
+                    var arraySegmentValue = (ArraySegment<byte>)data.Value;
+                    dataSegments.AddRange(arraySegmentValue.ToArray());
+                }
+                sbMessage = new SBMessage(dataSegments.ToArray());
             }
             else
             {
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
     using System.Collections;
     using System.Collections.Generic;
     using System.IO;
+    using System.Linq;
     using System.Runtime.Serialization;
     using Microsoft.Azure.Amqp;
     using Microsoft.Azure.Amqp.Encoding;
@@ -171,6 +172,14 @@ namespace Microsoft.Azure.ServiceBus.Amqp
             {
                 var byteArrayValue = (byte[])amqpMessage.ValueBody.Value;
                 sbMessage = new SBMessage(byteArrayValue);
+            }
+            else if (amqpMessage.BodyType == SectionFlag.Data
+                && amqpMessage.DataBody != null
+                && amqpMessage.DataBody.Count() > 0)
+            {
+                var dataBody = amqpMessage.DataBody.FirstOrDefault().Value;
+                var dataBodyAsByteArray = (ArraySegment<byte>)dataBody;
+                sbMessage = new SBMessage(dataBodyAsByteArray.ToArray());
             }
             else
             {
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
@@ -168,17 +168,18 @@ namespace Microsoft.Azure.ServiceBus.Amqp
             }
 
             SBMessage sbMessage;
-            if (amqpMessage.BodyType == SectionFlag.AmqpValue && amqpMessage.ValueBody.Value != null)
+
+            if ((amqpMessage.BodyType & SectionFlag.AmqpValue) != 0
+                && amqpMessage.ValueBody.Value != null)
             {
                 var byteArrayValue = (byte[])amqpMessage.ValueBody.Value;
                 sbMessage = new SBMessage(byteArrayValue);
             }
-            else if (amqpMessage.BodyType == SectionFlag.Data
-                && amqpMessage.DataBody != null
-                && amqpMessage.DataBody.Count() > 0)
+            else if ((amqpMessage.BodyType & SectionFlag.Data) != 0
+                && amqpMessage.DataBody?.Count() > 0)
             {
                 var dataSegments = new List<byte>();
-                foreach (Data data in amqpMessage.DataBody)
+                foreach (var data in amqpMessage.DataBody)
                 {
                     var arraySegmentValue = (ArraySegment<byte>)data.Value;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
```diff
@@ -172,6 +173,14 @@ public static SBMessage AmqpMessageToSBMessage(AmqpMessage amqpMessage)
                 var byteArrayValue = (byte[])amqpMessage.ValueBody.Value;
                 sbMessage = new SBMessage(byteArrayValue);
             }
+            else if (amqpMessage.BodyType == SectionFlag.Data
+                && amqpMessage.DataBody != null
+                && amqpMessage.DataBody.Count() > 0)
+            {
+                var dataBody = amqpMessage.DataBody.FirstOrDefault().Value;
```
jtaubensee: Would we ever need more thant he first DataBody? If so, what would be the best way to concatenate them?

jtaubensee:
We should also add a unit test for this, using the underlying Azure AMQP library.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
```diff
@@ -172,6 +173,14 @@ public static SBMessage AmqpMessageToSBMessage(AmqpMessage amqpMessage)
                 var byteArrayValue = (byte[])amqpMessage.ValueBody.Value;
                 sbMessage = new SBMessage(byteArrayValue);
             }
+            else if (amqpMessage.BodyType == SectionFlag.Data
```
nemakam: SectionFlag is over type [Flags]. So you will have to check that bit and not check for equality of entire object.
i.e., instead of `amqpMessage.BodyType == SectionFlag.Data`, you will have to do `(amqpMessage.BodyType & SectionFlag.AmqpValue) != 0`

nemakam: Similarly, change it in the previous line as well

nemakam: Similarly, you need to take care of SectionFlag.AmqpSequence

src/Microsoft.Azure.ServiceBus/Resources.resx
```diff
@@ -201,4 +201,7 @@
   <data name="ArgumentMustBePositive" xml:space="preserve">
     <value>The value of the argument {0} must be positive.</value>
   </data>
+  <data name="InvalidAmqpMessageProperty" xml:space="preserve">
+    <value>{0} is not a supported user property type.</value>
```
nemakam: We need to remember to add this to documentation, as to which all types we support as of today.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
```diff
@@ -167,11 +168,24 @@ public static SBMessage AmqpMessageToSBMessage(AmqpMessage amqpMessage)
...
-            if (amqpMessage.BodyType == SectionFlag.AmqpValue && amqpMessage.ValueBody.Value != null)
+
+            if ((amqpMessage.BodyType & SectionFlag.AmqpValue) != 0
+                && amqpMessage.ValueBody.Value != null)
             {
                 var byteArrayValue = (byte[])amqpMessage.ValueBody.Value;
                 sbMessage = new SBMessage(byteArrayValue);
             }
+            else if ((amqpMessage.BodyType & SectionFlag.Data) != 0
+                && amqpMessage.DataBody?.Count() > 0)
```
vinaysurya: >DataBody?.Count() [](start = 31, length = 17)

Avoid Linq where possible, just add a check to DataBody! = null., here. the Ienumerable will will make sure that it is entered only if count > 0

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
```diff
@@ -167,11 +168,24 @@ public static SBMessage AmqpMessageToSBMessage(AmqpMessage amqpMessage)
...
+            {
+                var dataSegments = new List<byte>();
+                foreach (var data in amqpMessage.DataBody)
+                {
+                    var arraySegmentValue = (ArraySegment<byte>)data.Value;
+                    dataSegments.AddRange(arraySegmentValue.ToArray());
```
vinaysurya: >.ToArray() [](start = 59, length = 10)

why toArray() here?.. Is is not sufficient to add the values directly here?

src/Microsoft.Azure.ServiceBus/Resources.resx
jtaubensee: Created #119 to track this.

bmmathe:
Is there something specific I need to add to my SubscriptionClient or RegisterMessageHandler code to get this to work? I'm still getting a null Body even though in the ServiceBus Explorer I can see the message body with no issues.  I'm currently using version 2.0 of this library.

SeanFeldman:
@bmmathe could you share the code you're using to send/receive?

bmmathe:
```csharp
internal class ServiceBusHelper
    {
        private static CancellationTokenSource cancelSource = new CancellationTokenSource();

        private static List<SubscriptionClient> _clients;
        public static async Task MainAsync(string serviceBusConnectionString)
        {
            _clients = new List<SubscriptionClient>();
            var client1 = new SubscriptionClient(serviceBusConnectionString, "topicfolder/topic1", "sub1", ReceiveMode.PeekLock);
            var client2 = new SubscriptionClient(serviceBusConnectionString, "topicfolder/topic2", "sub2", ReceiveMode.PeekLock);
            _clients.Add(client1);
            _clients.Add(client2);

            client1.RegisterMessageHandler(
                async (message, token) =>
                {
                    try
                    {
                        var body = Encoding.UTF8.GetString(message.Body);
                        var myMessage = JsonConvert.DeserializeObject<MyMessage>(body);
                        //do stuff
                        await client1.CompleteAsync(message.SystemProperties.LockToken);
                    }
                    catch (Exception ex)
                    {
                        Log.LogException(ex);
                    }
                },
                new MessageHandlerOptions()
                    {MaxConcurrentCalls = 1, AutoComplete = false});

            client2.RegisterMessageHandler(
                async (message, token) =>
                {
                    try
                    {
                        var body = Encoding.UTF8.GetString(message.Body);
                        var myMessage = JsonConvert.DeserializeObject<MyMessage>(body);
                        //do stuff
                        await client2.CompleteAsync(message.SystemProperties.LockToken);
                    }
                    catch (Exception ex)
                    {
                        Log.LogException(ex);
                    }
                },
                new MessageHandlerOptions()
                { MaxConcurrentCalls = 1, AutoComplete = false });

            cancelSource.Token.WaitHandle.WaitOne();

            // Close the client after the ReceiveMessages method has exited.
            foreach (var client in _clients)
            {
                await client.CloseAsync();
            }
            Log.LogMessage("ReceiveMessages closed.", MessageLevel.Info);
        }                
    }
```

SeanFeldman:
This doesn't look right. With client 2.0 you have to provide exception callback into `MessageHandlerOptions` constructor.

SeanFeldman:
Just tried your code with Microsoft.Azure.ServiceBus client 2.0.0 sending messages using ServiceBus Explorer 4.0.105 and it worked fine. Messages were sent using `Stream`. Probably would be better to share your repro project as a whole to see what's going on.

bmmathe:
The sender in my case for the "old" messages wasn't ServiceBus Explorer.  We had been using the ServiceBus API to send messages via HTTP calls.  The API version we were referencing was api-version=2013-08.  It seems as though the messages created from this API version are incompatible with the 2.0.0 .Net Standard client.  It's not a huge deal since we moved off of calling the SB API back in May, just strange that 0.0.5-preview could process the messages but not the 2.0.0 version.

Issue #35, opened 2016-12-14
OnSession Handler


jtaubensee:
@vinaysurya to provide details

jtaubensee:
Removed help-wanted label until details are provided

jtaubensee:
Closing based on #126

PR #126, opened 2017-04-11
On session feature

vinaysurya:
## Description
This codeflow implements the OnSessions Feature For DotNet Core library. Also noticed some issues with the way we were doing close on our Client Entities, have addressed them as well.

Addresses:
https://github.com/Azure/azure-service-bus-dotnet/issues/35

- [X ] The code builds without any errors.


src/Microsoft.Azure.ServiceBus/Amqp/AmqpLinkCreator.cs
```diff
@@ -46,19 +46,33 @@ protected AmqpLinkCreator(string entityPath, ServiceBusConnection serviceBusConn
...
+            {
+                // Create Link
                 AmqpObject link = this.OnCreateAmqpLink(connection, this.amqpLinkSettings, session);
                 await link.OpenAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
                 return link;
             }
-            catch (Exception)
+            catch (Exception exception)
             {
-                session?.Abort();
```
nemakam: Why remove this line?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -379,13 +379,13 @@ public Task<IList<Message>> PeekAsync(int maxMessageCount)
...
         public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handler)
         {
-            this.RegisterMessageHandler(handler, new RegisterHandlerOptions() { ReceiveTimeOut = this.OperationTimeout });
+            this.RegisterMessageHandler(handler, new RegisterMessageHandlerOptions());
         }
 
-        public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handler, RegisterHandlerOptions registerHandlerOptions)
+        public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handler, RegisterMessageHandlerOptions registerMessageHandlerOptions)
         {
-            registerHandlerOptions.ReceiveTimeOut = this.OperationTimeout;
```
nemakam: if ReceiveTimeout is not specified in the options, then we will have to add it ourselves

src/Microsoft.Azure.ServiceBus/Amqp/AmqpLinkCreator.cs
vinaysurya: I have not removed it. Just seperated the session creation/open from the link creation and not aborting the session when link creation fails

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: the default is initialized in the constructor for RegisterMessageHandlerOptions

PR #146, opened 2017-04-26
Exposing MessageSender/Receiver

jtaubensee:
## Description
Addresses #57.
This PR publicly exposes the MessageSender/Receiver to provide more advanced functionality is not found in the Queue/Topic/Subscription clients.

The following changes are a part of this PR:
* Publicly exposed MessageSender and Receiver
  * Replaced original `MessageReceiver` with `AmqpMessageReceiver`, and renamed it to `MessageReceiver`
  * Replaced original `MessageSender` with `AmqpMessageSender`, and renamed it to `MessageSender`
  * Replaced original `MessageSession` with `AmqpSessionClient`, and renamed it to `MessageSession`
* Unsealed Queue/Topic/Subscription clients

Previously, things were setup such that we would be able to implement different transport mechanisms utilizing an abstract MessageSender/Receiver. There are no plans for this in the foreseeable future, so we have removed some of the unneeded complexity.


src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -27,11 +128,13 @@ protected MessageReceiver(ReceiveMode receiveMode, TimeSpan operationTimeout, Re
...
 
-        public abstract string Path { get; }
-
         public ReceiveMode ReceiveMode { get; protected set; }
 
-        public virtual int PrefetchCount
+        /// <summary>
+        /// Get Prefetch Count configured on the Receiver.
+        /// </summary>
+        /// <value>The upper limit of events this receiver will actively receive regardless of whether a receive operation is pending.</value>
```
nemakam: >events [](start = 38, length = 6)

messages?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -8,16 +8,117 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            ReceiveMode receiveMode = ReceiveMode.PeekLock,
+            RetryPolicy retryPolicy = null)
+            : base(nameof(MessageSender) + StringUtility.GetRandomString(), retryPolicy ?? RetryPolicy.Default)
+        {
+            this.ReceiveMode = receiveMode;
+            this.operationTimeout = serviceBusConnection.OperationTimeout;
+            this.messageReceivePumpSyncLock = new object();
+            this.entityPath = entityPath;
+            this.ServiceBusConnection = serviceBusConnection;
+            this.PrefetchCount = prefetchCount;
```
nemakam: >this.PrefetchCount = prefetchCount; [](start = 12, length = 35)

this doesnt seem right. you are not taking prefetchCount in the constructor

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -8,16 +8,117 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(serviceBusConnection.SasKeyName, serviceBusConnection.SasKey);
+            this.CbsTokenProvider = new TokenProviderAdapter(tokenProvider, serviceBusConnection.OperationTimeout);
+            this.ReceiveLinkManager = new FaultTolerantAmqpObject<ReceivingAmqpLink>(this.CreateLinkAsync, this.CloseSession);
+            this.RequestResponseLinkManager = new FaultTolerantAmqpObject<RequestResponseAmqpLink>(this.CreateRequestResponseLinkAsync, this.CloseRequestResponseSession);
+            this.requestResponseLockedMessages = new ConcurrentExpiringSet<Guid>();
+        }
+
+        internal MessageReceiver(
```
nemakam: >internal MessageReceiver( [](start = 8, length = 25)

there seems to be a disconnect  between the public and the internal constructors. 
You can have multiple constructors, but all of them should essentially reach to the same constructor which initializes all properties of the class.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -8,16 +8,117 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            string entityPath,
+            MessagingEntityType? entityType,
+            ReceiveMode receiveMode,
+            int prefetchCount,
+            ServiceBusConnection serviceBusConnection,
+            ICbsTokenProvider cbsTokenProvider,
+            string sessionId,
+            RetryPolicy retryPolicy,
+            bool isSessionReceiver = false)
+            : base(nameof(MessageReceiver) + StringUtility.GetRandomString(), retryPolicy ?? RetryPolicy.Default)
```
nemakam: >: base(nameof(MessageReceiver) + StringUtility.GetRandomString(), retryPolicy ?? RetryPolicy.Default) [](start = 12, length = 101)

as commented earlier, let all the constructors finally reach this constructor and then this could do `: base()`

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -67,26 +175,35 @@ internal set
             }
         }
 
+        public virtual string Path
+        {
+            get { return this.entityPath; }
```
nemakam: >entityPath [](start = 30, length = 10)

you dont need this. you could convert this to autoProperty with get and private set

nemakam: Similarly for others

---
In reply to: [113600980](https://github.com/Azure/azure-service-bus-dotnet/pull/146#discussion_r113600980) [](ancestors = 113600980)

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -388,21 +505,432 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
             this.OnMessageHandlerAsync(messageHandlerOptions, handler).GetAwaiter().GetResult();
         }
 
-        protected abstract Task<IList<Message>> OnReceiveAsync(int maxMessageCount, TimeSpan serverWaitTime);
+        public override async Task OnClosingAsync()
```
nemakam: >OnClosingAsync [](start = 35, length = 14)

not completely related to your change, but could you please make this `OnClosingAsync()` protected EVERYWHERE?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,91 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+                throw Fx.Exception.ArgumentNullOrWhiteSpace(entityPath);
+            }
+
+            this.ownsConnection = true;
+        }
+
+        public MessageSender(string entityPath, ServiceBusConnection serviceBusConnection, RetryPolicy retryPolicy = null)
+            : base(nameof(MessageSender) + StringUtility.GetRandomString(), retryPolicy ?? RetryPolicy.Default)
+        {
+            this.OperationTimeout = serviceBusConnection.OperationTimeout;
```
nemakam: >his.OperationTimeout = serviceBusConnection.OperationTimeout; [](start = 13, length = 61)

like MessageReceiver, have only one constructor do entire initialization. ALl other constructors should just drive to that one particular overload

nemakam:
>     sealed class AmqpSessionClient : IMessageSessionEntity

we might have to expose this as well

---
Refers to: src/Microsoft.Azure.ServiceBus/Amqp/AmqpSessionClient.cs:11 in e3c83fb. [](commit_id = e3c83fbc4325d1bca10c58c6d705c1636fec32b2, deletion_comment = False)

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageSession.cs
```diff
@@ -8,7 +8,6 @@ namespace Microsoft.Azure.ServiceBus.Amqp
     using System.Threading.Tasks;
     using Azure.Amqp;
     using Core;
-    using Messaging.Amqp;
 
     class AmqpMessageSession : MessageSession
```
nemakam: >AmqpMessageSession [](start = 10, length = 18)

we may have to expose this as well

nemakam: and rename it to MessageSession 

---
In reply to: [113602592](https://github.com/Azure/azure-service-bus-dotnet/pull/146#discussion_r113602592) [](ancestors = 113602592)

nemakam:
:clock1:

jtaubensee:
I still have some more feedback to address

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageSession.cs
jtaubensee: You're saying for the sake of thoroughness, to follow the same pattern with `MessageSession` and `AmqpMessageSession`? Where we remove `AmqpMessageSession` and add it's functionality to `MessageSession`? Just double checking...

nemakam: Yes. More specifically, we need to expose a way to use "advanced functionality" for sessions as MessageReceiver doesn't handle sessions. 

---
In reply to: [113775192](https://github.com/Azure/azure-service-bus-dotnet/pull/146#discussion_r113775192) [](ancestors = 113775192)

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -45,49 +44,29 @@ protected ServiceBusConnection(TimeSpan operationTimeout, RetryPolicy retryPolic
...
-
         internal FaultTolerantAmqpObject<AmqpConnection> ConnectionManager { get; set; }
 
         public Task CloseAsync()
         {
             return this.ConnectionManager.CloseAsync();
         }
 
-        internal MessageSender CreateMessageSender(string entityPath)
+        internal IMessageSender CreateMessageSender(string entityPath)
```
vinaysurya: >CreateMessageSender [](start = 32, length = 19)

Now that we are making Sender and Receiver Public and having explicit constructors to create them, we can kill this method of creating senders/receivers from here?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -8,30 +8,128 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            int prefetchCount = DefaultPrefetchCount,
+            ReceiveMode receiveMode = ReceiveMode.PeekLock,
+            RetryPolicy retryPolicy = null)
+            : this(entityPath, null, receiveMode, serviceBusConnection, null, retryPolicy)
+        {
+            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(serviceBusConnection.SasKeyName, serviceBusConnection.SasKey);
+            this.CbsTokenProvider = new TokenProviderAdapter(tokenProvider, serviceBusConnection.OperationTimeout);
+        }
+
+        internal MessageReceiver(
```
nemakam: >MessageReceiver [](start = 17, length = 15)

There seems to be a lot of internal constructors. can you see if you can either minimize them to one. Removing all of them would be the best option, but depends on how we are using them in the tests. If tests can be edited to use public constructors, it would be perfect.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -44,12 +142,16 @@ public virtual int PrefetchCount
...
+                ReceivingAmqpLink link;
+                if (this.ReceiveLinkManager.TryGetOpenedObject(out link))
+                {
+                    link.SetTotalLinkCredit((uint)value, true, true);
+                }
             }
         }
 
-        public virtual long LastPeekedSequenceNumber
+        public long LastPeekedSequenceNumber
```
nemakam: >public [](start = 8, length = 6)

wondering if this should be made protected and not public

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -67,26 +169,23 @@ internal set
...
-        }
+        public virtual string Path { get; private set; }
+
+        public DateTime LockedUntilUtc { get; protected set; }
+
+        public string SessionId { get; protected set; }
+
+        internal TimeSpan OperationTimeout { get; private set; }
 
         internal MessagingEntityType? EntityType { get; set; }
```
nemakam: >set [](start = 56, length = 3)

Updating the entityType after creating the link doesnt work, we should mark the `set` private.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,98 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+    using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Azure.Amqp;
+    using Microsoft.Azure.Amqp.Encoding;
+    using Microsoft.Azure.Amqp.Framing;
+    using Microsoft.Azure.ServiceBus.Amqp;
+    using Microsoft.Azure.ServiceBus.Primitives;
 
-    internal abstract class MessageSender : ClientEntity, IMessageSender
+    public sealed class MessageSender : ClientEntity, IMessageSender
```
nemakam: >sealed [](start = 11, length = 6)

MessageReceiver is not sealed. We should maintain consistency between the two. IMO, we should not have them sealed

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,98 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            RetryPolicy retryPolicy = null)
+            : this(connectionStringBuilder.GetNamespaceConnectionString(), connectionStringBuilder.EntityPath, retryPolicy)
+        {
+        }
+
+        public MessageSender(
+            string connectionString, 
+            string entityPath, 
+            RetryPolicy retryPolicy = null)
+            : this(entityPath, new ServiceBusNamespaceConnection(connectionString), retryPolicy ?? RetryPolicy.Default)
```
nemakam: > ?? RetryPolicy.Default [](start = 95, length = 23)

we can remove this default initialization as it is already been done elsewhere

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,98 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            this.ServiceBusConnection = serviceBusConnection;
+            this.CbsTokenProvider = cbsTokenProvider;
+            this.SendLinkManager = new FaultTolerantAmqpObject<SendingAmqpLink>(this.CreateLinkAsync, this.CloseSession);
+            this.RequestResponseLinkManager = new FaultTolerantAmqpObject<RequestResponseAmqpLink>(this.CreateRequestResponseLinkAsync, this.CloseRequestResponseSession);
         }
 
         internal TimeSpan OperationTimeout { get; }
 
-        protected MessagingEntityType? EntityType { get; set; }
+        internal MessagingEntityType? EntityType { get; set; }
```
nemakam: >set [](start = 56, length = 3)

private set

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,98 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+
+        string Path { get; }
```
nemakam: >string Path { get; } [](start = 8, length = 20)

public virtual?

src/Microsoft.Azure.ServiceBus/Amqp/AmqpSubscriptionClient.cs
```diff
@@ -8,10 +8,10 @@ namespace Microsoft.Azure.ServiceBus.Amqp
     using Azure.Amqp;
     using Core;
     using Filters;
-    using Messaging.Amqp;
 
     internal sealed class AmqpSubscriptionClient : IInnerSubscriptionClient
```
nemakam: >internal sealed [](start = 4, length = 15)

I think we should try to add a new sample which clearly mentions that its for advanced scenarios and use MessageReceivers and sessions etc just to validate. 
You could take it up as a separate PR, but it will help you close on few things like whether this class should be internal/public, or whether the class needs to be renamed

src/Microsoft.Azure.ServiceBus/Amqp/AmqpSubscriptionClient.cs
```diff
@@ -62,11 +64,14 @@ public MessageReceiver InnerReceiver
         /// <value>The number of messages that the subscription client can simultaneously request.</value>
         public int PrefetchCount
         {
-            get => this.ServiceBusConnection.PrefetchCount;
-
+            get => this.prefetchCount;
```
nemakam: >get => this.prefetchCount; [](start = 12, length = 26)

this doesnt need to maintain its own prefetch count. It could just pass the value to innerReceiver and calls innerReceiver.Prefetch count for get.

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -89,11 +90,14 @@ public SubscriptionClient(string connectionString, string topicPath, string subs
...
         {
-            get => this.ServiceBusConnection.PrefetchCount;
-
+            get => this.prefetchCount;
             set
             {
-                this.ServiceBusConnection.PrefetchCount = value;
+                if (value < 0)
+                {
+                    throw Fx.Exception.ArgumentOutOfRange(nameof(this.PrefetchCount), value, "Value must be greater than 0");
```
nemakam: >Value must be greater than 0 [](start = 94, length = 28)

value cannot be negative. it can be 0

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
jtaubensee: We define this in the `IMessageReceiver` interface. Whats your thought for making it protected?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
jtaubensee: Left a single internal constructor, and unit tests all use public constructors now

jtaubensee:
@nemakam / @vinaysurya - Ok, I think it have either addressed all feedback, or responded with a comment. Assuming this next build passes, will you take another peek and sign off if all looks well?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -5,23 +5,98 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+            string entityPath,
+            ServiceBusConnection serviceBusConnection,
+            RetryPolicy retryPolicy = null)
+            : this(entityPath, null, serviceBusConnection, null, retryPolicy)
+        {
+            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(
+                serviceBusConnection.SasKeyName,
+                serviceBusConnection.SasKey);
+            this.CbsTokenProvider = new TokenProviderAdapter(tokenProvider, serviceBusConnection.OperationTimeout);
+        }
```
vinaysurya: Since we do not have tests that test connection behavior with re-using connections etc, may be we can remove this as well. We can add back this variant, when we have more tests etc for sharing connections

vinaysurya:
Other than that one comment about removing an additional constructor, the rest looks good John! Thanks for addressing this!

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
jtaubensee: So, it looks like this in the test: 
```csharp
var connection = new ServiceBusNamespaceConnection(TestUtility.NamespaceConnectionString);
var sender = new MessageSender(queueName, connection);
var receiver = new MessageReceiver(queueName, connection, receiveMode: ReceiveMode.PeekLock);
```

Issue #134, opened 2017-04-18
Connection pooling for clients


jtaubensee:
## Description
This issue is to track the implementation of a connection pool for the Queue/Topic/Subscription clients.

This feature would enable the following scenarios:
* Latency optimization (shared connections)
* Throughput optimization (multiple connections)

Other things to keep in mind:
* Connections would be returned to global pool when the client is closed/disposed
* Connection pool would be per namespace

PR #152, opened 2017-05-04
Removed Sender/Receiver constructor for ServiceBusNamespaceConnection

jtaubensee:
## Description
Removing constructor that accepts `ServiceBusNamespaceConnection` for `MessageSender` and `MessageReceiver`. We don't want to expose functionality like this until we have connection pooling ironed out. See #134 for details.

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
@@ -7,7 +7,6 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
     using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.Azure.ServiceBus.Core;
-    using Microsoft.Azure.ServiceBus.Primitives;
     using Xunit;
 
     public class ExpectedMessagingExceptionTests
@@ -16,10 +15,9 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
         async Task MessageLockLostExceptionTest()
         {
             const int messageCount = 2;
-            var connection = new ServiceBusNamespaceConnection(TestUtility.NamespaceConnectionString);
 
-            var sender = new MessageSender(TestConstants.NonPartitionedQueueName, connection);
-            var receiver = new MessageReceiver(TestConstants.NonPartitionedQueueName, connection, receiveMode: ReceiveMode.PeekLock);
+            var sender = new MessageSender(TestUtility.NamespaceConnectionString, TestConstants.NonPartitionedQueueName);
+            var receiver = new MessageReceiver(TestUtility.NamespaceConnectionString, TestConstants.NonPartitionedQueueName, receiveMode: ReceiveMode.PeekLock);
 
             try
             {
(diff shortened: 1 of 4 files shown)


nemakam:
>             string connectionString,

not related, but could you do some input validation in constructor? In this as well as other constructor?

---
Refers to: src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs:42 in e66fcdd. [](commit_id = e66fcddd92346c4bf0af4bcf924dc315ce60065b, deletion_comment = False)

nemakam:
>             : this(connectionStringBuilder.GetNamespaceConnectionString(), connectionStringBuilder.EntityPath, receiveMode, retryPolicy, prefetchCount)

like sean mentioned, you may wanna change this to `connectionStringBuilder?`

---
Refers to: src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs:37 in d4cd0e2. [](commit_id = d4cd0e2e7416f5ed9df8dcc031379e8fa28be5b1, deletion_comment = False)

nemakam:
>             : this(connectionStringBuilder.GetNamespaceConnectionString(), connectionStringBuilder.EntityPath, receiveMode, retryPolicy, prefetchCount)

ad other params

---
In reply to: [301558462](https://github.com/Azure/azure-service-bus-dotnet/pull/152#issuecomment-301558462) [](ancestors = 301558462)

---
Refers to: src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs:37 in d4cd0e2. [](commit_id = d4cd0e2e7416f5ed9df8dcc031379e8fa28be5b1, deletion_comment = False)

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
```diff
@@ -48,66 +46,11 @@ public class ExpectedMessagingExceptionTests
             }
         }
 
-        // TODO: Update this when Session is implemented
-        /*
-        [Fact]
-        async Task SessionLockLostExceptionTest()
```
nemakam: revert this. Let it stay commented for now. We need to change this to work with receiver instead of queueClient

nemakam:
:shipit:

jtaubensee:
So to be clear, the null check becomes relevant when we use this:
`: this(connectionStringBuilder?.GetNamespaceConnectionString()`

SeanFeldman:
If `connectionStringBuilder` is null, should it throw? We rely on `connectionStringBuilder` to provide connection string and path. If it's null, then can sender/receiver operate as normal @nemakam ?

nemakam:
@SeanFeldman , the idea is not that we continue working without the object. We currently fail with nullRefException when we try `connectionStringBuilding.GetNamespaceConnectionString()`. Instead, we will now end up passing null and throw `ArgumentNullException` in the next step which seems to be a better response.

SeanFeldman:
Right. It will fail with the `ArgumentNullException`. I was implying that it would be nicer to let know that the  connection string builder was null, to set the focus of the developer to the right spot, but because we chain constructors, this is not an option. Thanks @nemakam

Issue #106, opened 2017-03-17
Extensibility API


SeanFeldman:
Allow extension API to override/enrich certain behaviors 
Good candidates could be:
- Message ID generator (currently "baked" into the library) - #24 Allow registration of message id factory
- Mutation of incoming and outgoing message
- etc

This will open up an opportunity to plug-in customization or provide those as independent packages for consumption (@jtaubensee brought up an idea of a `KeyVault` package to secure messages, see example below).

Here's a proposed API:

`IClientEntity` to expose a property `Extensions` of a type `IExtension`

```c#
public interface IExtension
{
  IExtension MessageIdGenerator(Func<Message>, string> idGenerator);
  IExtension IncomingMessageMutator(Func<Message, Task<Message>> mutator);
  IExtension OutgoingMessageMutator(Func<Message, Task<Message>> mutator);
  // etc
}
```

This will allow configuration of customization that are only needed in the following manner:

```c#
queueClient.Extensions.MessageIdGenerator(message => $"{message.Id}_{DateTime.Now.ToString("s")}");
queueClient.Extensions.OutgoingMessageMutator(message =>HeaderMutator(message));
```

It will also allow creation of extension methods that could be packaged and distributed:

```c#

// KeVault package 

namespace Microsoft.Azure.ServiceBus.Extensions.KeyVault
{
  static class KeVaultExtensionExtensions
  {
    static void UseKeyVault(this IExtension extention, params...)
    {
       extention.IncomingMessageMutator(funcIn);
       extention.OutgoingMessageMutator(funcOut);
    }
  }
}
//...
```

Usage (of the package and in-project extensions):

```c#
queueClient.Extensions.UseKeyVault(params);
queueClient.Extensions.MessageIdGenerator(message => $"{message.Id}_{DateTime.Now.ToString("s")}");
queueClient.Extensions.OutgoingMessageMutator(message =>HeaderMutator(message));
```

**Main points**

1. Allow very specific extension points with clear API.
1. Extension points should not be "getting in the way" on the root level (ie only whoever is interested would access the API by going through `Extensions`).
1. Allow registration of multiple extensions for any extension APIs (e.g. multiple outgoing message mutators).

**Drawbacks**

Extensions are currently provided during the execution phase, which could lead to accidental dynamic changes by user. To solve this, it would require to split into two phases: (1) configuration phase and (2) execution phase, when extension points are locked. To implement 2-phased approach it would require to review the general API so that 

```c# 
[...]

PR #153, opened 2017-05-05
Adding plugin support

jtaubensee:
## Description
Addresses #106.

Adds initial plugin points for all clients / sender receivers. This will be updated once the design in finalized.


jtaubensee:
This is still very much a work in progress, but wanted to keep the ball rolling.

Here are some things that still need thought:
* Exception handling - I was thinking we could do something like [`OnMessageOptions.ExceptionReceived`](https://docs.microsoft.com/dotnet/api/microsoft.servicebus.messaging.onmessageoptions.exceptionreceived?view=azureservicebus-4.0.0)
  * Should we allow you to implement a custom exception that would allow us to send the message anyway? For example, if we were using KeyVault, we wouldn't want to send an unencrypted message, but if we were just doing something with performance counters, we'd still like to send, even if our custom logic failed.
* What other extension points are needed? Right now, I have:
  * BeforeMessageSend
  * AfterMessageReceive
* When can the plugins be registered? https://github.com/Azure/azure-service-bus-dotnet/issues/106#issuecomment-298720362

SeanFeldman:
> Should we allow you to implement a custom exception that would allow us to send the message anyway?

Custom exception or exception controlling callback that allows to prevent exception from bubbling up?

> When can the plugins be registered?

I would advocate at configuration time only.

> What other extension points are needed?

If you treat it as a pipeline, perhaps should have `Before` and 'After`? For the outgoing  pipeline you could have before and after, that would be used to stamp the brokered message before it hits the wire.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -211,24 +214,44 @@ public Task<IList<Message>> ReceiveAsync(int maxMessageCount)
...
+                }
+                else
+                {
+                    processedMessageList = new List<Message>();
+                    foreach (var message in unprocessedMessageList)
+                    {
+                        var processedMessage = message;
+                        foreach (var afterEachMessageReceiveTask in this.AfterEachMessageReceiveTasks)
+                        {
+                            processedMessage = await afterEachMessageReceiveTask(message);
```
SeanFeldman: configureawait(false)

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -110,10 +113,30 @@ public Task SendAsync(Message message)
...
+                }
+                else
+                {
+                    processedMessageList = new List<Message>();
+                    foreach (var message in messageList)
+                    {
+                        var processedMessage = message;
+                        foreach (var beforeEachMessageSendTask in this.BeforeEachMessageSendTasks)
+                        {
+                            processedMessage = await beforeEachMessageSendTask(message);
```
SeanFeldman: configureawait(false)

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -148,10 +171,19 @@ public Task SendAsync(Message message)
 
             try
             {
+                var processedMessage = message;
+                if (this.BeforeEachMessageSendTasks != null)
+                {
+                    foreach (var beforeEachMessageSendTask in this.BeforeEachMessageSendTasks)
+                    {
+                        processedMessage = await beforeEachMessageSendTask(message);
```
SeanFeldman: configureawait(false)

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -102,6 +103,8 @@ protected MessageReceiver(ReceiveMode receiveMode, TimeSpan operationTimeout, Re
             this.messageReceivePumpSyncLock = new object();
         }
 
+        public IList<Func<Message, Task<Message>>> AfterEachMessageReceiveTasks { get; set; }
```
SeanFeldman: public setter?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -71,8 +71,11 @@ public class MessageSender : ClientEntity, IMessageSender
             this.CbsTokenProvider = cbsTokenProvider;
             this.SendLinkManager = new FaultTolerantAmqpObject<SendingAmqpLink>(this.CreateLinkAsync, this.CloseSession);
             this.RequestResponseLinkManager = new FaultTolerantAmqpObject<RequestResponseAmqpLink>(this.CreateRequestResponseLinkAsync, this.CloseRequestResponseSession);
+            this.BeforeEachMessageSendTasks = new List<Func<Message, Task<Message>>>();
         }
 
+        public IList<Func<Message, Task<Message>>> BeforeEachMessageSendTasks { get; set; }
```
SeanFeldman: public setter?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -211,24 +214,44 @@ public Task<IList<Message>> ReceiveAsync(int maxMessageCount)
...
+
                 await this.RetryPolicy.RunOperation(
                     async () =>
                     {
-                        messages = await this.OnReceiveAsync(maxMessageCount, serverWaitTime).ConfigureAwait(false);
+                        unprocessedMessageList = await this.OnReceiveAsync(maxMessageCount, serverWaitTime).ConfigureAwait(false);
                     }, serverWaitTime)
                     .ConfigureAwait(false);
+
+                if (this.AfterEachMessageReceiveTasks == null)
```
SeanFeldman: How about adding a default noop task and skip the evaluation?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -110,10 +113,30 @@ public Task SendAsync(Message message)
 
             try
             {
+                IList<Message> processedMessageList;
+
+                if (this.BeforeEachMessageSendTasks == null)
```
SeanFeldman: same idea with noop

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
jtaubensee: @SeanFeldman - I think a no-op makes sense here, still working on the best way to do it though.

jtaubensee:
Just added an ExceptionHandler for `ServiceBusPlugin`. This probably still needs a few things:
* Do we want to have a custom exception, or just reuse the `ServiceBusException.IsTransient` logic? I'm guessing we want something else
* Do we want to reuse `ExceptionReceivedEventArgs` from `MessageHandlerOptions`?

jtaubensee:
Regarding exception handling, any exception in a plugin will be thrown to the corresponding action. So you will get an exception on `QueueClient.SendAsync`, unless you throw a `ServiceBusPluginException` with `ShouldCompleteOperation` set to true (in your plugin code). The idea here is that, you wouldnt want to send an unencrypted message if you can't reach KeyVault, but you might be ok with the message not being compressed.

Outstanding items to be addressed:
* Only registering plugins at startup. @SeanFeldman - Were you thinking this would only be accessible through the constructor then?
* Additional extension points

@nemakam / @vinaysurya please take a look as well.

SeanFeldman:
@jtaubensee registration - the thought was to discourage registration of plugins during runtime. With no clear configuration and execution phases, don't see how it would be possible other than during construction.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -960,5 +1009,48 @@ void ThrowIfNotPeekLockMode()
...
+                this.RegisteredPlugins = new List<ServiceBusPlugin>();
+            }
+            else if (this.RegisteredPlugins.Any(p => p.GetType() == serviceBusPlugin.GetType()))
+            {
+                throw new ArgumentException(nameof(serviceBusPlugin), Resources.PluginAlreadyRegistered.FormatForUser(nameof(serviceBusPlugin)));
+            }
+            this.RegisteredPlugins.Add(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(ServiceBusPlugin serviceBusPlugin)
```
SeanFeldman: What would be the case to unregister a plugin? Curious to understand.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,47 @@ public class MessageSender : ClientEntity, IMessageSender
             }
         }
 
+        private async Task<Message> ProcessMessages(Message message)
```
SeanFeldman: Method name `ProcessMessages`, but it only processes a single message.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,47 @@ public class MessageSender : ClientEntity, IMessageSender
...
+            }
+
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    processedMessage = await plugin.BeforeMessageSend(message).ConfigureAwait(false);
+                }
+                catch (Exception e)
```
SeanFeldman: Same exception filtering as above could be used.

src/Microsoft.Azure.ServiceBus/QueueClient.cs
```diff
@@ -305,5 +305,23 @@ public Task CancelScheduledMessageAsync(long sequenceNumber)
...
+            this.InnerReceiver.RegisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(ServiceBusPlugin serviceBusPlugin)
+        {
+            this.InnerSender.UnregisterPlugin(serviceBusPlugin);
+            this.InnerReceiver.UnregisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(Type serviceBusPluginType)
```
SeanFeldman: Is this overload needed?

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -308,5 +308,20 @@ public Task AddRuleAsync(string ruleName, Filter filter)
...
+        {
+            this.InnerSubscriptionClient.InnerReceiver.RegisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(ServiceBusPlugin serviceBusPlugin)
+        {
+            this.InnerSubscriptionClient.InnerReceiver.UnregisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(Type serviceBusPluginType)
```
SeanFeldman: Same question, is this overload needed?

src/Microsoft.Azure.ServiceBus/TopicClient.cs
```diff
@@ -142,5 +142,20 @@ public Task CancelScheduledMessageAsync(long sequenceNumber)
...
+        {
+            this.InnerSender.RegisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(ServiceBusPlugin serviceBusPlugin)
+        {
+            this.InnerSender.UnregisterPlugin(serviceBusPlugin);
+        }
+
+        public void UnregisterPlugin(Type serviceBusPluginType)
```
SeanFeldman: same as above

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
jtaubensee: This might be a stretch, and maybe that means its not needed for now, but I can think of a scenario where I would want to unregister when I continually receive errors with my plugin, and I want to continue using the same client...

SeanFeldman: > This might be a stretch, and maybe that means its not needed for now

Then don't add it 😃 

>  I can think of a scenario where I would want to unregister when I continually receive errors with my plugin

I'm not sure about this. If you have a component that you rely on and it misbehaves, you would find the issue and redeploy the component. I might be off on this one, but feels like adding this w/o actual need would be premature.

jtaubensee: I can get behind that. Any thoughts @nemakem / @vinaysurya?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,40 @@ internal set
...
+                catch (Exception ex) when (ex is ServiceBusPluginException pluginException && pluginException.ShouldCompleteOperation)
+                {
+                    continue;
+                }
+            }
+            return processedMessage;
+        }
+
+        private async Task<IList<Message>> ProcessMessages(IList<Message> messageList)
+        {
```
nemakam: return right here if there are no registered plugins. need to iterate through all the messages.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,40 @@ internal set
...
+            {
+                return message;
+            }
+
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    processedMessage = await plugin.AfterMessageReceive(message).ConfigureAwait(false);
```
nemakam: >processedMessage = await plugin.AfterMessageReceive(message).ConfigureAwait(false); [](start = 20, length = 83)

add log statements before and after calling plugin

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,40 @@ internal set
...
+                }
+                catch (Exception ex) when (ex is ServiceBusPluginException pluginException && pluginException.ShouldCompleteOperation)
+                {
+                    continue;
```
nemakam: >continue; [](start = 20, length = 9)

should log here as well

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
nemakam: I partly agree with @Sean here. I think we should be able to register plugin only during creation of Queue unless we see a strong reason to change things dynamically.
But at the same time, we might want to have `UnregisterPlugin` or `ClosePlugin` equivalent so that plugin can probably release some resources of its own if required. But this would require us to expose additional methods on plugin. 
So in effect, we might end up having two more functions `OnPluginRegistered` and `OnPluginUnregistered` methods in `ServiceBusPlugin`

---
In reply to: [117356385](https://github.com/Azure/azure-service-bus-dotnet/pull/153#discussion_r117356385) [](ancestors = 117356385)

nemakam: But now to think of it, i can think of scenarios where we may want to unregister plugin. 
Lets say we have a plugin to tackle inter-op scenarios for users to switch from old client to new client. At some point in time they realise that they are done with the interop handling (upgrade scenarios), and now want to unregister the plugin. 

---
In reply to: [117356385](https://github.com/Azure/azure-service-bus-dotnet/pull/153#discussion_r117356385) [](ancestors = 117356385)

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,40 @@ public class MessageSender : ClientEntity, IMessageSender
...
+            {
+                return message;
+            }
+
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    processedMessage = await plugin.BeforeMessageSend(message).ConfigureAwait(false);
```
nemakam: >processedMessage = await plugin.BeforeMessageSend(message).ConfigureAwait(false); [](start = 20, length = 81)

like in receiver, log before and after calling the plugin to debug.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,40 @@ public class MessageSender : ClientEntity, IMessageSender
...
+                }
+                catch (Exception ex) when (ex is ServiceBusPluginException pluginException && pluginException.ShouldCompleteOperation)
+                {
+                    continue;
+                }
+            }
+            return processedMessage;
+        }
+
+        private async Task<IList<Message>> ProcessMessages(IList<Message> messageList)
```
nemakam: >private async Task<IList<Message>> ProcessMessages(IList<Message> messageList) [](start = 8, length = 78)

return quick if no plugins found instead of iterating through the messages

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,40 @@ public class MessageSender : ClientEntity, IMessageSender
...
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    processedMessage = await plugin.BeforeMessageSend(message).ConfigureAwait(false);
+                }
+                catch (Exception ex) when (ex is ServiceBusPluginException pluginException && pluginException.ShouldCompleteOperation)
+                {
+                    continue;
```
nemakam: >continue; [](start = 20, length = 9)

like in receiver, log the exception

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,40 @@ public class MessageSender : ClientEntity, IMessageSender
...
+            }
+
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    processedMessage = await plugin.BeforeMessageSend(message).ConfigureAwait(false);
+                }
+                catch (Exception ex) when (ex is ServiceBusPluginException pluginException && pluginException.ShouldCompleteOperation)
```
nemakam: >pluginException.ShouldCompleteOperation [](start = 94, length = 39)

What are your opinions on exposing this on the plugin? Who should decide whether this is an exception that impacts or not? - User or the plugin developer?

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusPluginException.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.Azure.ServiceBus
+{
+    using System;
+
+    /// <summary>
+    /// Exception for ServiceBus plugin errors.
+    /// </summary>
+    public class ServiceBusPluginException : Exception
+    {
+        public ServiceBusPluginException(bool shouldCompleteOperation)
```
nemakam: >shouldCompleteOperation [](start = 46, length = 23)

maybe rename to `shouldContinueOperation`?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
SeanFeldman: My gut feeling tells me the user. Plugin is there to do the job, not to decide on the overall course of actions.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
SeanFeldman: Interesting, saw a comment from @nemakam, but now it's gone.
Reason I'm against this method for now is YAGNI. If at some point in time a plugin is no longer needed, it should be removed, not just disabled.

nemakam: Yeah, I cant find my own comment. Reposting it here:

>I partly agree with @Sean here. I think we should be able to register plugin only during creation of Queue unless we see a strong reason to change things dynamically. But at the same time, we might want to have UnregisterPlugin or ClosePlugin equivalent so that plugin can probably release some resources of its own if required. But this would require us to expose additional methods on plugin. So in effect, we might end up having two more functions OnPluginRegistered and OnPluginUnregistered methods in ServiceBusPlugin

>But now to think of it, i can think of scenarios where we may want to unregister plugin. Lets say we have a plugin to tackle inter-op scenarios for users to switch from old client to new client. At some point in time they realise that they are done with the interop handling (upgrade scenarios), and now want to unregister the plugin.

nemakam: And regarding YAGNI, I think we might end up working on building an interop plugin very soon. Given that, I feel we could leave this UnRegister as-is and not remove it as we don't see any issues with it other than YAGNI.

>  If at some point in time a plugin is no longer needed, it should be removed, not just disabled.
Based on current implementation, IMO  `UnregisterPlugin` seems to be "removing plugin" and not just "disabling" it.

SeanFeldman:
Agree that UnregisterPlugin would achieve the same result, but why would
one register a plugin to unregister it shortly after if plugin is not
needed :D​

jtaubensee:
@SeanFeldman / @nemakam - Ok, I think I've incorporated most of your feedback now. Here are some points to clarify:
* I've removed the `ServiceBusPluginException`, and added a bool to the `ServiceBusPlugin`
  * `public virtual bool ShouldContinueOnException => false;` - this will handle whether or not the error should continue or throw
* I've also added an abstract name that is required. This allows us to identify plugins easier;
  * `public abstract string Name { get; }`
* Added logging

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,51 @@ internal set
...
+            {
+                return message;
+            }
+
+            var processedMessage = message;
+            foreach (var plugin in this.RegisteredPlugins)
+            {
+                try
+                {
+                    MessagingEventSource.Log.PluginCalled(plugin.Name, message.MessageId);
```
SeanFeldman: `PluginCallStarted` & `PluginCallCompleted`?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,51 @@ internal set
...
+                    processedMessage = await plugin.AfterMessageReceive(message).ConfigureAwait(false);
+                    MessagingEventSource.Log.PluginCompleted(plugin.Name, message.MessageId);
+                }
+                catch (Exception ex)
+                {
+                    MessagingEventSource.Log.PluginException(plugin.Name, message.MessageId, ex.Message);
```
SeanFeldman: `PluginCallFailed` or `PluginCallThrewException`?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -960,5 +1011,39 @@ void ThrowIfNotPeekLockMode()
...
+            {
+                return;
+            }
+            if (serviceBusPluginName == null)
+            {
+                throw new ArgumentNullException(nameof(serviceBusPluginName), Resources.ArgumentNullOrWhiteSpace.FormatForUser(nameof(serviceBusPluginName)));
+            }
+            if (this.RegisteredPlugins.Any(p => p.Name == serviceBusPluginName))
+            {
+                var plugin = this.RegisteredPlugins.First(p => p.Name == serviceBusPluginName);
```
SeanFeldman: The way I interpret this is that you can have more than one plugin of different types with the same name. That doesn't sound right.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -960,5 +1011,39 @@ void ThrowIfNotPeekLockMode()
 
             MessagingEventSource.Log.RegisterOnMessageHandlerStop(this.ClientId);
         }
+
+        public void RegisterPlugin(ServiceBusPlugin serviceBusPlugin)
```
SeanFeldman: What about name that is already used?

src/Microsoft.Azure.ServiceBus/Core/ServiceBusPlugin.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Core
+{
+    using System.Threading.Tasks;
+
+    public abstract class ServiceBusPlugin
+    {
+        public abstract string Name { get; }
```
SeanFeldman: Should it default to something like a new Guid?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
jtaubensee: Good catch - it should be this: `this.RegisteredPlugins.Any(p => p.Name == serviceBusPlugin.Name)`

The name should be unique, so if you start using our KeyVault plugin the name/id will be set by the plugin itself, so you won't be able to register it multiple times.

src/Microsoft.Azure.ServiceBus/Core/ServiceBusPlugin.cs
jtaubensee: I would expect that this be set by the plugin, so we wouldnt want to auto generate this. We are hoping to prevent users from calling use multiple times with the same plugin.

jtaubensee:
@SeanFeldman - 
> I don't see plugin work been done differently based on the fact if it's a sender or a receiver

They are largely the same however, the one difference is the extension point that is called.
* Sender: `plugin.BeforeMessageSend`
* Receiver: `plugin.AfterMessageReceive`

SeanFeldman:
Right @jtaubensee, the invocation takes place at different places, but the registration logic remains the same. That's why I felt the configuration part could be the same.

SeanFeldman:
PS: i can raise a PR tomorrow. Would that work?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -201,6 +199,46 @@ internal set
...
+                        throw;
+                    }
+                }
+            }
+            return processedMessage;
+        }
+
+        private async Task<IList<Message>> ProcessMessages(IList<Message> messageList)
+        {
+            if (!this.RegisteredPlugins.Any())
```
nemakam: >.Any() [](start = 39, length = 6)

Use RegisteredPlugins.Count instead of Any() since this is in the hot-path and Any() is considerably slower than Count

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -160,6 +162,51 @@ internal set
...
+                        throw;
+                    }
+                }
+            }
+            return processedMessage;
+        }
+
+        private async Task<IList<Message>> ProcessMessages(IList<Message> messageList)
+        {
+            if (this.RegisteredPlugins == null || !this.RegisteredPlugins.Any())
```
danielmarbach: Can this ever be null? Wouldn't be more efficient to make sure we always have a list there and use the count property instead of linq?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -90,6 +92,51 @@ public class MessageSender : ClientEntity, IMessageSender
...
+                        throw;
+                    }
+                }
+            }
+            return processedMessage;
+        }
+
+        private async Task<IList<Message>> ProcessMessages(IList<Message> messageList)
+        {
+            if (this.RegisteredPlugins == null || !this.RegisteredPlugins.Any())
```
danielmarbach: Same question as above

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
SeanFeldman: [Can't be null](https://github.com/Azure/azure-service-bus-dotnet/blob/6cc56bda017d28737b1eae47124cd6d115bc2086/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L132).

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
SeanFeldman: Same answer with 👍 for `IList<T>.Count`

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: @danielmarbach, looks like this is already corrected in the checked code
https://github.com/Azure/azure-service-bus-dotnet/blob/6cc56bda017d28737b1eae47124cd6d115bc2086/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L227

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
vinaysurya: Same here, looks already taken care of in current code

PR #161, opened 2017-05-20
Small cleanups

SeanFeldman:
Some cleanups that are possible with C# 7

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessagingUtilities.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessagingUtilities.cs
@@ -25,8 +25,8 @@ namespace Microsoft.Azure.ServiceBus
 
         public static bool ShouldRetry(Exception exception)
         {
-            ServiceBusException serviceBusException = exception as ServiceBusException;
-            return serviceBusException != null && serviceBusException.IsTransient;
+            var serviceBusException = exception as ServiceBusException;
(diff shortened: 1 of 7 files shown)


SeanFeldman:
@jtaubensee targeting your branch so that you can pick it up and continue if you like it.

src/Microsoft.Azure.ServiceBus/MessagingUtilities.cs
```diff
@@ -25,8 +25,8 @@ public static TimeSpan CalculateRenewAfterDuration(DateTime lockedUntilUtc)
 
         public static bool ShouldRetry(Exception exception)
         {
-            ServiceBusException serviceBusException = exception as ServiceBusException;
-            return serviceBusException != null && serviceBusException.IsTransient;
+            var serviceBusException = exception as ServiceBusException;
+            return serviceBusException?.IsTransient == true;
```
nemakam: >?.IsTransient == true; [](start = 38, length = 22)

Is `== true` required?

nemakam:
:shipit:

src/Microsoft.Azure.ServiceBus/MessagingUtilities.cs
SeanFeldman: Just adds a bit of clarity, but definitely not required.

PR #162, opened 2017-05-20
Remove unnecessary continue statements

SeanFeldman:
no need in continue as we're not re-throwing


SeanFeldman:
@jtaubensee targeting your branch

PR #163, opened 2017-05-20
Remove null check for registered plugins collection

SeanFeldman:
Simplify code by replacing checks in several places with initialized collection.


SeanFeldman:
@jtaubensee targeting your branch

PR #170, opened 2017-06-01
Fix ApiApprover to list public API properly

### LINKED ISSUE #169: APIApprover: Missing public members
The file [here](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt), seems to only have public interfaces and classes listed and not the members inside of those. @seanfeldman seems to think there is a setting that needs to be flipped.

SeanFeldman:
A bit anecdotal, but this library didn't expect to have MSFT code to use it.
The check on the namespace prevents any types that are from `System` and `Microsoft` namespaces. 
This project is in the `Microsoft.Azure.ServiceBus` namespace, which should be accepted.

This PR fixes #169


SeanFeldman:
@jtaubensee it's ready

Issue #174, opened 2017-06-05
ExceptionReceivedEventArgs.Action 


SeanFeldman:
Possible values for `ExceptionReceivedEventArgs.Action` are not documented and appear to be free text in [`MessageReceivePump`](https://github.com/Azure/azure-service-bus-dotnet/blob/63713319375c2b34cb8e71c5f8a07f6494a30324/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs) and duplicated in [`SessionReceivePump`](https://github.com/Azure/azure-service-bus-dotnet/blob/63713319375c2b34cb8e71c5f8a07f6494a30324/src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs).

Suggest to 
1. Extract those into constants
1. Document --> related issue https://github.com/Azure/azure-service-bus/issues/27

SeanFeldman:
Also, the two classes use inconsistent strings to represent similar actions (`UserCallback` vs `User Callback Exception`).

PR #176, opened 2017-06-05
ExceptionReceivedEventArgs.Action consolidation

SeanFeldman:
Fixes issue #174 

@nemakam or vinaysurya please review

@sethmanheim these are the actions for the **new client**, not sure about the old/current one

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
+++ b/src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
@@ -1,0 +1,13 @@
+﻿namespace Microsoft.Azure.ServiceBus
+{
+    static class ExceptionReceivedEventArgsAction
+    {
+        public const string Complete = "Complete";
+        public const string Abandon = "Abandon";
+        public const string UserCallback = "UserCallback";
+        public const string Receive = "Receive";
+        public const string RenewLock = "RenewLock";
+        public const string AcceptMessageSession = "AcceptMessageSession";
+        public const string CloseMessageSession = "CloseMessageSession";
+    }
+}
--- a/src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
+++ b/src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
@@ -5,16 +5,16 @@
     /// <summary>Action taking place when <see cref="ExceptionReceivedEventArgs"/> is raised for a processed message.</summary>
     public static class ExceptionReceivedEventArgsAction
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Microsoft.Azure.ServiceBus
+{
+    static class ExceptionReceivedEventArgsAction
```
nemakam: what are your thoughts on making this public?

SeanFeldman: I assume you're asking me @nemakam. I'm on a fence about that. What would be the a strong use case for having it public?

nemakam: I'm sure it doesn't fit into the mainline scenario. But even for a very simple use case wherein a customer wants to keep a count of exceptions occurred during each operation, he would need to segregate all the exceptions received based on this action.

SeanFeldman:
@nemakam made it public. Might need help with XML doc to describe better those actions.
image

src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Microsoft.Azure.ServiceBus
+{
+    using Microsoft.Azure.ServiceBus.Primitives;
+
+    /// <summary>Action taking place when <see cref="ExceptionReceivedEventArgs"/> is raised for a processed message.</summary>
```
nemakam: > is raised for a processed message.

This doesn't sound right. We have actions here like `AcceptMessageSession` and `CloseMessageSession` and `Receive`

SeanFeldman: Ha-ha, hence [the comment](https://github.com/Azure/azure-service-bus-dotnet/pull/176#issuecomment-306372568): 

> Might need help with XML doc to describe better those actions.

SeanFeldman: btw, I tweaked it a little. Will let the team come up with a better xml summaries.

nemakam: Okay, but meanwhile could you please remove the 'processed message' part? Incomplete documentation isn't as bad as wrong documentation :)

SeanFeldman:
@nemakam done

PR #179, opened 2017-06-11
Clarifying what management library should be used

SeanFeldman:
@binzywu @jtaubensee the document is referring to `Microsoft.Azure.Management.ServiceBus` but that one is older than `Microsoft.Azure.Management.ServiceBus.Fluent`. In case I'm wrong, please clarify readme.md and close this PR. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Once you have completed the above, you can run `dotnet test` from the `/test/Mic
 
 ### Can I manage Service Bus entities with this library?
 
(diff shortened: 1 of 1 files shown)


SeanFeldman:
Looks like `Microsoft.Azure.Management.ServiceBus` is the library to use.
Issue is with version. Sample uses outdated version 0.0.1, where current version of the library is 0.2.0.
This sample needs an update 😃

Issue #177, opened 2017-06-09
QueueClient.RegisterMessageHandler -- no way to timeout if no messages are received


CodeCharm:
Maybe I'm approaching this wrong? I'm a little new to this library; please direct me to what docs/posts I missed if I am thinking about this the wrong way.

`QueueClient.RegisterMessageHandler` blocks until a message is received. It's not awaitable; it doesn't return a `Task`.

I would like to establish a timeout in case there are no messages to receive. But it seems to drop into the message pump and doesn't return unless I have AutoComplete = true, or set the CancellationToken from within the loop, neither of which helps if I never get a message.

Consequently, there's no simple way to timeout that I can see. I'll have to wrap the call in a `Task.Run`, then use a `Task.WhenAny` with a `CancellationToken` (or something, I don't know yet). It feels wrong. Especially because the inner `MessageReceiver` uses a Task in `OnMessageHandlerAsync` while waiting for the message to get to the `MessageReceiverPump`.

I'm suggesting this be added as a feature, and I'm willing to try to make a contribution to do it, but I don't want to waste anyone's time.

My thought was to use the `ReceiveTimeOut` property in the `MessageHandlerOptions`, and use a `CancellationTokenSource` in the `OnMessageHandlerSync` to break out after the delay, but then throw a `TimeoutException` if it was the delay that broke out of the awaiter.

SeanFeldman:
`QueueClient.RegisterMessageHandler` is a message pump. It's purpose is to continuously run and invoke a handler provided by you for each incoming message, not process a single message and be done.

It sounds what you want to do is to receive a single message with a timeout. For that you have [`QueueClient.ReceiveAsync(TimeSpan serverWaitTime)`](https://github.com/Azure/azure-service-bus-dotnet/blob/6cc56bda017d28737b1eae47124cd6d115bc2086/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L256). Is that what you're after?

nemakam:
Every thing that @SeanFeldman mentioned is right, except for a small typo...
The `ReceiveAsync` is a method on `MessageReceiver` and not `QueueClient`. (The link is right though).

Closing this issue.

CodeCharm:
No, I wanted a message pump. It's just that I have no external control over when to end the message pump. Once I'm in it, I'm stuck in it until a message is received.

[...]

Issue #154, opened 2017-05-14
RegisterMessageHandler takes exactly 1 minute to complete


jack4it:
## Actual Behavior
1. The call of RegisterMessageHandler takes exactly 1 minute to complete

## Expected Behavior
1. Should this not be faster?
2. Why 1 minute though?

## Versions  
- OS platform and version: win 10, server 2012/2016
- .NET Version: 1.1.x
- NuGet package version or commit ID: 0.0.2, 0.0.3

vinaysurya:
It does not have to be. I think today we wait for a whole minute waiting for a message to show up while we register the handler for the first time. We could just try waiting for a lesser time or for no time and still go ahead and register the handler anyway. I'll make that change and submit a PR.

MPOSEIDON:
I am having the same issue. Why it is taking 1 minute to register message handler?

SeanFeldman:
@MPOSEIDON see PR #183

nemakam:
#183 is checked in. Next release should solve this issue

PR #183, opened 2017-06-14
RegisterMessageHandler should not block for 1 minute if there are no messages

SeanFeldman:
Address #154 / #177 by reducing initial message receive wait time in the `MessageReceivePump` from 60 seconds to 1 second.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Constants.cs
+++ b/src/Microsoft.Azure.ServiceBus/Constants.cs
@@ -40,5 +40,7 @@ namespace Microsoft.Azure.ServiceBus
         public static readonly int DefaultClientPumpPrefetchCount = 5;
 
         public static readonly TimeSpan NoMessageBackoffTimeSpan = TimeSpan.FromSeconds(5);
+
+        public static readonly TimeSpan MessageReceiverStartPumpInitialReceiveTimeout = TimeSpan.Zero;
     }
 }
--- a/src/Microsoft.Azure.ServiceBus/Constants.cs
+++ b/src/Microsoft.Azure.ServiceBus/Constants.cs
@@ -41,6 +41,6 @@ namespace Microsoft.Azure.ServiceBus
 
         public static readonly TimeSpan NoMessageBackoffTimeSpan = TimeSpan.FromSeconds(5);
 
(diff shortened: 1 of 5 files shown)


nemakam:
@SeanFeldman 
Regarding testing, I don't think this exact bug requires its own test case, but having this bug also shows that we don't have a test case which registers the message handler before sending the message. IMO, having one such test is important. So could you please add this test case which will take care of this scenario?

SeanFeldman:
@nemakam ready for review

src/Microsoft.Azure.ServiceBus/Constants.cs
```diff
@@ -40,5 +40,7 @@ static class Constants
         public static readonly int DefaultClientPumpPrefetchCount = 5;
 
         public static readonly TimeSpan NoMessageBackoffTimeSpan = TimeSpan.FromSeconds(5);
+
+        public static readonly TimeSpan MessageReceiverStartPumpInitialReceiveTimeout = TimeSpan.FromSeconds(1);
```
nemakam: @SeanFeldman Why 1?
0 serverWaitTime seems to be more ideal

nemakam: 0 server wait time returns immediately if there are no messages waiting currently.

SeanFeldman: I tried that with `TimeSpan.Zero` and it wasn't the case, it was take much longer than 10 seconds. I also tried milliseconds, same effect.

SeanFeldman:
Rebased on the latest `dev`

src/Microsoft.Azure.ServiceBus/Constants.cs
nemakam: Oh I see. Let me see what that issue is. 
I have a feeling 1 second might be too small. But I don't have a better suggestion as well. Lets leave it at 1 second for now and i'll update it if we come up with a better value.

SeanFeldman: 👍 perhaps we should raise an issue about this so that at some point in time we can have a look at it and figure out what's going on?

SeanFeldman:
@nemakam  I've raised a separate issue #189 for the `serverWaitTime`

PR #184, opened 2017-06-14
Changing Amqp message body type from AmqpValue to AmqpData

### LINKED ISSUE #138: Unable to send "pure" body content
## Actual Behavior
1. `MemoryStream` (with simple `foo bar` text) is converted to `byte[]` via `ToArray()` and sent via ServiceBus
2. Real content of message is binary-serialized data like `@base64Binar3http://schemas.microsoft.com/2003/10/Serialization/?  ?foo bar?`

## Expected Behavior
1. Send "pure" byte content as message payload, for compatibility with other platforms and old (legacy) clients

## Explanation

We use ServiceBus for interaction between applications written on .NET and Java, where message content is XML text, written as "pure" bytes in to `Stream` of `BrokeredMessage`. We are planning to introduce new services written in NETCore, which should talk with existing services.

There is no problem to call `memoryStream.ToArray()` to send `byte[]` (in new services), but it's unacceptable that this data is being "packed" to some kind of serialization-envelope, because this require additional de-serialization step on receiving.

I wrote [sample app](https://github.com/justdmitry/ServiceBusStreamInterop) with both `WindowsAzure.ServiceBus` and `Microsoft.Azure.ServiceBus`, demonstrating the problem, here is output:

sample

How can I send "pure" content in `Message`? I found #98 with explanation about removing `Stream` body content, and agree about "need to be in memory" and "sends are taking a long time, when the latency is in fact caused by the reading of the stream", but is it really needed to use serialization-envelope, while `byte[]` is the only content/payload type?

## Versions  
- `Microsoft.Azure.ServiceBus 0.0.2-preview` and `Microsoft.Azure.ServiceBus 0.0.3-preview`
- `WindowsAzure.ServiceBus 4.0.0`

binzywu:
## Description
Changing Message.Body to ArraySegment and changing the amqp message body type to Data for the interoperate with old .Net client library

This resolves #138 

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Message.cs
+++ b/src/Microsoft.Azure.ServiceBus/Message.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.ServiceBus
         /// Creates a new Message
         /// </summary>
         public Message()
-            : this(default(byte[]))
+            : this(default(ArraySegment<byte>))
         {
         }
 
--- a/src/Microsoft.Azure.ServiceBus/Message.cs
+++ b/src/Microsoft.Azure.ServiceBus/Message.cs
@@ -30,10 +30,18 @@ namespace Microsoft.Azure.ServiceBus
         /// <summary>
         /// Creates a new message from the specified payload.
         /// </summary>
-        /// <param name="array"></param>
-        public Message(byte[] array)
+        /// <param name="body"></param>
+        public Message(byte[] body) : this(new ArraySegment<byte>(body))
         {
-            this.Body = array;
+        }
+
+        /// <summary>
+        /// Creates a new message from the specified payload.
+        /// </summary>
+        /// <param name="body"></param>  
+        public Message(ArraySegment<byte> body)
+        {
+            this.Body = body;
             this.SystemProperties = new SystemPropertiesCollection();
             this.UserProperties = new Dictionary<string, object>();
         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -44,10 +52,10 @@ public Message(byte[] array)
         /// <remarks>
         /// The easiest way to create a new message from a string is the following:
         /// <code>
-        /// message.Body = System.Text.Encoding.UTF8.GetBytes("Message1");
+        /// message.Body = new ArrySegment&lt;byte&gt;(System.Text.Encoding.UTF8.GetBytes("Message1"));
         /// </code>
         /// </remarks>
-        public byte[] Body { get; set; }
+        public ArraySegment<byte> Body { get; set; }
```
SeanFeldman: This feels off. To create a new message with payload, ctor accepts `byte[]`.
To assign a body later I need to use `ArraySegment<byte>`. Why can't it be `byte[]` all the way?

binzywu: so let's move byte[] ctor? 😃

binzywu: Yes, this is a breaking change. I am not saying we should do this. 😃 Comments are needed to see what's the right move here. ArraySegment can certainly bring many values and even we don't expose it as Body, internally we will still use ArraySegment as there are optimizations can be done. But public interface could remain as byte[]. We are very close to make this library GA for it's first version and want to get some critical decisions done before it goes 1.0.

SeanFeldman: > so let's move byte[] ctor?

Not sure I follow. Did you mean "remove"?

> Yes, this is a breaking change

Well, this is a pre-release 🙂 


`ArraySegment` - not arguing. Probably internally makes a lot of sense. Externally perhaps should be `byte[]` as the first/default option.

binzywu: K. Thanks for the feedback. Yeah, I will have another discussion with team tomorrow and will update the pull request. But generally this change is mainly to change the message to use Amqp Data as we do some extra processing in both old client and service for AmqpValue and Data will be processed as is which is desired.

PR #187, opened 2017-06-15
Auto code cleanup

nemakam:
## Description
Ran the VS-Resharper auto code cleanup tool

This checklist is used to make sure that common guidelines for a pull request are followed.

- [x] **I have read the [contribution guidelines](./CONTRIBUTING.md).**
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR.
- [x] The pull request does not introduce breaking changes (unless a major version change occurs in the assembly and module).
- [x] If applicable, the public code is properly documented.
- [x] Pull request includes test coverage for the included changes.
- [x] The code builds without any errors.

PR #188, opened 2017-06-15
Replacing System.TimeoutException with ServiceBusTimeoutException

### LINKED ISSUE #173: Wrap TimeoutException in ServiceBusTimeoutException with IsTransient as true
[Documentation](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-exceptions) on `TimeoutException` says:

> Retry might help in some cases; add retry logic to code.

This is a challenge today because `TimeoutException` is [not derived from the `ServiceBusException`](https://github.com/Azure/azure-service-bus-dotnet/blob/784fc14cf348c472b5a25eb5aa8e2afc8b43a56d/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs#L97) which exposes `IsTransient`. The translation is using 

Should the client throw a custom (specific to this library) timeout exception derived from `MessagingException`? 

To provide more context, with the WinsowsAzure.ServiceBus client there are a few issues when handling exceptions with `OnMessageOptions.ExceptionReceived` and trying to determine if a timeout exception is transient or not (issue https://github.com/Azure/azure-service-bus/issues/28 and https://github.com/Azure/azure-service-bus/issues/27).

SeanFeldman:
Fixes #173 

Introduces `ServiceBusTimeoutException` to replace `System.TimeoutException` that cannot be marked as transient.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
@@ -193,7 +193,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
                 case OperationCanceledException _:
                     return new ServiceBusException(true, message, exception);
 
-                case TimeoutException _ when referenceId != null:
+                case TimeoutException _:
                     return new ServiceBusTimeoutException(message, exception);
             }
 
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.Azure.ServiceBus.Primitives;
+
 namespace Microsoft.Azure.ServiceBus.Amqp
 {
     using System;
@@ -192,7 +194,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
                     return new ServiceBusException(true, message, exception);
 
                 case TimeoutException _ when referenceId != null:
-                    return new TimeoutException(message, exception);
+                    return new ServiceBusTimeoutException(message, exception);
             }
 
             return exception;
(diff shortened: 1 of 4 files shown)


SeanFeldman:
@nemakam are tests still not executed on external PRs?

nemakam:
@SeanFeldman , nope. I don't know if they ever will due to secrets management. Do you have a suggestion on how to run external PRs without the risk of exposing the secrets?

SeanFeldman:
Good question. Let's raise an issue for that as well. Might be in backlog for a while, but worth keeping around.

nemakam:
To think of it, I don't know if the implementation going to be as straight forward as this. What about actual connection level `TimeoutException` which isn't thrown from the backend? Not sure how to test that though...

nemakam:
I'm going to keep this PR open for a while. Let me also do some manual testing on it...

SeanFeldman:
👍 
I'm running the test suite right now.
It's still marked as WIP. Definitely need to cover all the options. Please let me know what I've missed.

SeanFeldman:
@nemakam I'm noticing some flakiness in the tests. When executed in bulk, some fail. When executed individually, they pass.

nemakam:
@SeanFeldman , found the issue of flakiness. 😎 
The test `OnMessageRegistrationWithoutPendingMessagesReceiveAndDelete` that was recently added forgot to close `QueueClient` before leaving the test and the background pump which it had initiated in `RecieveAndDelete` mode ended up swallowing a lot of messages which were supposed to reach other test clients. Fixed it as part of #187 itself

SeanFeldman:
You are awesome!
Probably will make sense to wait for your PR and then I'll revise this on
top of the new Dev branch.

SeanFeldman:
@nemakam I've rebased this PR on the latest `dev` to incorporate changes in #191.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
```diff
@@ -192,7 +194,7 @@ public static Exception GetClientException(Exception exception, string reference
                     return new ServiceBusException(true, message, exception);
 
                 case TimeoutException _ when referenceId != null:
```
nemakam: This case is true only when referenceId!=null.
Cases when referenceId is null, we still end up throwing `TimeoutException` instead of `ServiceBusTimeoutException`. Please take care of that scenario.

vinaysurya: May be just remove that referenceid check on a timeout exception will be fine.

SeanFeldman: 👍 wasn't sure what that original code was for. Changed.

PR #193, opened 2017-06-21
Add namespace name to OnMessage exception

### LINKED ISSUE #94: Namespace name in the OnMessage exception handler
Creating a new issue from part of #34.

Verbatim from @seanfeldman:
>When OnMessageOptions.ExceptionReceived event handler is invoked, it doesn't provide information on what namespace it's using. Having that information in the new client implementation would help for custom fail-over implementation.

SeanFeldman:
Fixes #94 

@vinaysurya please review


SeanFeldman:
@jtaubensee you might be interested in this one. This fixes the left out issue from #34.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -176,7 +179,7 @@ void CancelAutoRenewlock(object state)
             }
             catch (Exception exception)
             {
-                this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.Complete));
+                this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.Complete, namespaceName));
```
vinaysurya: this. ?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1101,7 +1101,7 @@ void ThrowIfNotPeekLockMode()
                 }
 
                 this.receivePumpCancellationTokenSource = new CancellationTokenSource();
-                this.receivePump = new MessageReceivePump(this, registerHandlerOptions, callback, this.receivePumpCancellationTokenSource.Token);
+                this.receivePump = new MessageReceivePump(this, registerHandlerOptions, callback, ServiceBusConnection.NamespaceName, this.receivePumpCancellationTokenSource.Token);
```
vinaysurya: >ServiceBusConnection [](start = 98, length = 20)

Again, this. here

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -76,7 +79,7 @@ bool ShouldRenewLock()
                 catch (Exception exception)
                 {
                     MessagingEventSource.Log.MessageReceivePumpTaskException(this.messageReceiver.ClientId, string.Empty, exception);
-                    this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.Receive));
+                    this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.Receive, namespaceName));
```
nemakam: if we are passing namespace, I feel we should definitely pass the entity name as well. Could you include `messageReceiver.Path` as well

vinaysurya:
@SeanFeldman, thanks for taking this up! the changes look good to me. One thing  I was thinking though is would it be weird to expose only a namespace name on a entitylevel exception. What if someone comesback and says, I also need to know what entity name ran into the exception. So I was thinking how about exposing both as part of the args. that way users can have full control

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -15,28 +15,19 @@ sealed class SessionReceivePump
...
         readonly SemaphoreSlim maxPendingAcceptSessionsSemaphoreSlim;
 
-        public SessionReceivePump(
-            string clientId,
-            IMessageSessionEntity client,
-            ReceiveMode receiveMode,
-            SessionHandlerOptions sessionHandlerOptions,
-            Func<IMessageSession, Message, CancellationToken, Task> callback,
-            CancellationToken token)
+        public SessionReceivePump(string clientId, IMessageSessionEntity client, ReceiveMode receiveMode, SessionHandlerOptions sessionHandlerOptions, Func<IMessageSession, Message, CancellationToken, Task> callback, string namespaceName, CancellationToken token)
```
nemakam: Could you revert this change? This is a long line and it is more readable with line breaks in between.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -323,7 +314,7 @@ void RaiseExceptionRecieved(Exception e, string action)
                     // Lets not bother user with this exception.
                     if (!(exception is TaskCanceledException))
                     {
-                        this.sessionHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.RenewLock));
+                        this.sessionHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.RenewLock, namespaceName));
```
nemakam: I think there are other places which you need to cover. For instance `CloseSessionIfNeededAsync()` method. Do take a second look at the usage of `RaiseExceptionReceived`

test/Microsoft.Azure.ServiceBus.UnitTests/Primitives/ServiceBusNamespaceConnectionTests.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.Azure.ServiceBus.UnitTests.Primitives
+{
+    using Xunit;
+    public class ServiceBusNamespaceConnectionTests
+    {
+        [Fact]
+        public void Can_return_namespace_name_for_connection_string()
+        {
+            var namespaceConnection = new ServiceBusNamespaceConnection("Endpoint=sb://seanfeldman-test.servicebus.windows.net/;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=7ry17m@yb31tw1llw0rk=");
```
nemakam: 😈

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
SeanFeldman: I really think `this.` is a noise and wish we'd move away from it 😞
But for this PR will do.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
SeanFeldman: In that case I'll introduce some Context object rather than multiple parameters. It will also be better from a public API perspective as signature on the method won't be considered broken if additions are made.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
SeanFeldman: `CloseSessionIfNeededAsync()` invokes `RaiseExceptionReceived()` which raises `ExceptionReceivedEventArgs` in a centralized manner, appending to the passed in information the namespace name @nemakam . Or am I missing something else?

test/Microsoft.Azure.ServiceBus.UnitTests/Primitives/ServiceBusNamespaceConnectionTests.cs
SeanFeldman: image

SeanFeldman:
@vinaysurya @nemakam I've incorporated all the feedback and implemented the `ExceptionReceivedContext`. Please review.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -12,20 +12,25 @@ namespace Microsoft.Azure.ServiceBus
...
+            Func<Message, CancellationToken, Task> callback, 
+            string namespaceName, 
+            string entityPath, 
             CancellationToken pumpCancellationToken)
         {
             this.messageReceiver = messageReceiver ?? throw new ArgumentNullException(nameof(messageReceiver));
             this.registerHandlerOptions = registerHandlerOptions;
             this.onMessageCallback = callback;
+            this.namespaceName = namespaceName;
+            this.entityPath = entityPath;
```
vinaysurya: >entityPath [](start = 17, length = 10)

this.messageReceiver exposes a 'Path' already. we could simply use that?

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -115,7 +120,7 @@ bool ShouldRenewLock()
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageReceiverPumpUserCallbackException(this.messageReceiver.ClientId, message, exception);
-                this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.UserCallback));
+                this.registerHandlerOptions.RaiseExceptionReceived(new ExceptionReceivedEventArgs(exception, ExceptionReceivedEventArgsAction.UserCallback, this.namespaceName, this.entityPath));
```
vinaysurya: >this.entityPath [](start = 176, length = 15)

this.messageReceiver.Path should work

src/Microsoft.Azure.ServiceBus/SessionPumpHost.cs
```diff
@@ -13,13 +13,17 @@ internal sealed class SessionPumpHost
         readonly object syncLock;
         SessionReceivePump sessionReceivePump;
         CancellationTokenSource sessionPumpCancellationTokenSource;
+        readonly string namespaceName;
+        readonly string entityPath;
```
vinaysurya: >entityPath [](start = 24, length = 10)

potentially add this to the IMessageSessionEntity interface like we do in the IReceiverClient.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
nemakam: Nope... Not missing anything 🥇

nemakam:
I'm just adding on-hold label for now so that we don't check this in. There are some small changes in the requirements of this task. 
We are planning to move out of 'namespace' model, to a complete 'endpoint' based connection. Watch out for a PR on this shortly that explains it.

nemakam:
@SeanFeldman , refer #196.
Based on that, you might have to update your changes to give out endpoint details instead of namespace name

SeanFeldman:
Thanks @nemakam for heads up. I'll watch that PR.

test/Microsoft.Azure.ServiceBus.UnitTests/Primitives/ServiceBusNamespaceConnectionTests.cs
nemakam: @SeanFeldman , keeping in view the public naming entities we use, could you rename seanfeldman to maybe 'contoso' or 'fabrikam'? Your pick!

SeanFeldman: My pick is `land-of-oz` 😃

SeanFeldman:
While #196 is brewing, have incorporated your feedback @vinaysurya @nemakam 
Blocked until #196 makes it into dev.

SeanFeldman:
@vinaysurya are you sure to approve this? It's still marked as `on hold` and needs to incorporate changes from #196

vinaysurya:
That is fine, well merge it only when Neerajs changes are checked in and
after your change on top of that.

SeanFeldman:
Ok. I'm used to a flow when approved PR is a PR that can be merged 🙂

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -100,5 +100,9 @@ static void CloseConnection(AmqpConnection connection)
 
             return connection;
         }
+
+        /// <summary>Namespace name used for <see cref="ServiceBusConnection"/>.</summary>
+        /// <value>Namespace name.</value>
+        internal string NamespaceName => this.Endpoint.Host.Split(new[] {'.'}, StringSplitOptions.RemoveEmptyEntries)[0];
```
nemakam: Based on #196, you'll have to change this to show the entire endpoint and not just the namespace... :)

SeanFeldman: So for a
```
"Endpoint=sb://land-of-oz.servicebus.windows.net/;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=7ry17m@yb31tw1llw0rk="
```

Exception context would surface `land-of-oz.servicebus.windows.net` @nemakam ?

nemakam: Yes. That's right.
We have public domains as well as separate national clouds. The domain name `servicebus.windows.net` should not be assumed.

SeanFeldman: @nemakam are those (national clouds) documented anywhere?

nemakam: http://lmgtfy.com/?q=microsoft+azure+national+clouds 😄

SeanFeldman: Haha, @nemakam lmgtfu was a good one, but it still doesn't answer if those domain names are documented as list somewhere.

nemakam: I don't know if different domain names are documented anywhere, but the domain names are definitely different. Also, having it as endpoint instead of namespace just makes it 1% more future-proof.. :)

SeanFeldman: Not questioning that. I just find it... peculiar that those domains are not documented. @sethmanheim any ideas if that's documented anywhere for ASB? Thank you.

SeanFeldman:
When creating a `ServiceBusNamespaceConnection` with 
```
Endpoint=sb://land-of-oz.servicebus.windows.net/;....
```
getting back for `Endpoint`
```
amqps://land-of-oz.servicebus.windows.net
```

`ServiceBusConnectionStringBuilder` hardcodes `amqps` as endpoint`s scheme. Is that correct?

nemakam:
@SeanFeldman , Endpoint is a Uri, so having a `UriScheme` (in this case `amqps`) is a requirement. Based on use-case, we need to fetch the required properties. So you could do `endpoint.Authority` to get `land-of-oz.servicebus.windows.net`

SeanFeldman:
@nemakam correct. My question is why `UriScheme` is replaced, not why we need it 🙂

nemakam:
@SeanFeldman Ideally we should be parsing it from connection string, but I'd say YAGNI. As of now, there is no plan of extending the library to run with anything other than amqp protocol.

SeanFeldman:
Reason I've asked @nemakam, aside from failing test, is because users that would register with a connection string that would contain `sb://` as their URI scheme would expect to see that value back. Instead, they'll be getting `amqps://` which will be confusing.

nemakam:
Fair point. I need to test and verify the effects of keeping the scheme as-is, and changing the scheme only when user doesn't mention the scheme. Will take a look at it later. 
And just a reminder, there's action item pending on this PR. You still need to change the namespace to `endpoint` everywhere.

SeanFeldman:
@nemakam correct. I'm in the mids, but my tests are failing due to the question I've raised with the URI schema 🙂

nemakam:
Lets not block this PR for the above concern. Please go ahead with the PR assuming the `amqps` uriScheme., and when the `Urischeme` is picked up, will update your test accordingly.

SeanFeldman:
Perhaps to raise an issue and reference this PR's discussion? That way
context can be tracked down when work on the new issue is started.

SeanFeldman:
@nemakam I've pushed the change.
Also raised an issue #197

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -12,20 +12,22 @@ namespace Microsoft.Azure.ServiceBus
     sealed class MessageReceivePump
     {
         readonly Func<Message, CancellationToken, Task> onMessageCallback;
+        readonly string namespaceName;
```
nemakam: rename this to endpoint

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
```diff
@@ -0,0 +1,29 @@
+namespace Microsoft.Azure.ServiceBus.Primitives
```
nemakam: Add  copyright

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
```diff
@@ -0,0 +1,29 @@
...
+    using System;
+
+    /// <summary>Context provided for <see cref="ExceptionReceivedEventArgs"/> exception raised by the client.</summary>
+    public class ExceptionReceivedContext
+    {
+        /// <summary>Initializes a new instance of the <see cref="ExceptionReceivedContext" /> class.</summary>
+        /// <param name="action">The action associated with the exception.</param>
+        /// <param name="endpoint">The endpoint associated with the exception.</param>
+        /// <param name="entityPath">The entity path associated with the exception.</param>
+        public ExceptionReceivedContext(string action, string endpoint, string entityPath)
```
nemakam: internal constructor?

src/Microsoft.Azure.ServiceBus/QueueClient.cs
```diff
@@ -197,7 +197,8 @@ internal SessionPumpHost SessionPumpHost
                             this.sessionPumpHost = new SessionPumpHost(
                                 this.ClientId,
                                 this.ReceiveMode,
-                                this.SessionClient);
+                                this.SessionClient,
+                                this.ServiceBusConnection.Endpoint.ToString());
```
nemakam: Should this be `Endpoint.ToString()` or `Endpoint.Authority`?
Former would give out `amqps://abcd.servicebus.windows.net:80`, and latter would be `abcd.servicebus.windows.net`.
I feel the latter is cleaner. 
This will also relieve the need of #197

nemakam: Same with `MessageReceiver` and other classes

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -15,28 +15,22 @@ sealed class SessionReceivePump
...
         readonly SemaphoreSlim maxPendingAcceptSessionsSemaphoreSlim;
 
-        public SessionReceivePump(
-            string clientId,
-            IMessageSessionEntity client,
-            ReceiveMode receiveMode,
-            SessionHandlerOptions sessionHandlerOptions,
-            Func<IMessageSession, Message, CancellationToken, Task> callback,
-            CancellationToken token)
+        public SessionReceivePump(string clientId, IMessageSessionEntity client, ReceiveMode receiveMode, SessionHandlerOptions sessionHandlerOptions,
```
nemakam: Could you revert this formatting and make it multiline again?

test/Microsoft.Azure.ServiceBus.UnitTests/Primitives/ServiceBusNamespaceConnectionTests.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Microsoft.Azure.ServiceBus.Primitives;
```
nemakam: copyrights

nemakam: Move using inside namespace

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
SeanFeldman: Need a bot to do so 😃

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
SeanFeldman: If someone want to perform testing and needs to construct the object to mock behavior, `internal` will get in a way. Do you still think it should be internal?

nemakam: makes sense. We could leave it `public`

src/Microsoft.Azure.ServiceBus/QueueClient.cs
SeanFeldman: I could do that. Would it be still called "endpoint"?

PR #196, opened 2017-06-22
ServiceBusConnectionStringBuilder accepts FQDN endpoint instead of namespace name.

nemakam:
## Description
1. `ServiceBusConnectionStringBuilder` no longer takes namespace and instead expects FQDN of endpoint
Currently we take namespace name, and format the namespace ourselves. In this, we assume the domain name is always `servicebus.windows.net`, which may not be true in all the cases. So instead, we require the customer to provide the endpoint details along with the domain name 

2. Minor formatting changes in `ServiceBusConnectionStringBuilder` to trim the `connectionString`

3. `ConnectionString` might contain additional unknown key value properties which might not be recognized by `ServiceBusConnectionStringBuilder`. In this case, keeping track of it in a separate dictionary instead of throwing.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Azure.ServiceBus
 
         protected void InitializeConnection(ServiceBusConnectionStringBuilder builder)
         {
-            this.Endpoint = builder.Endpoint;
+            this.Endpoint = new Uri(builder.Endpoint);
             this.SasKeyName = builder.SasKeyName;
             this.SasKey = builder.SasKey;
             this.ConnectionManager = new FaultTolerantAmqpObject<AmqpConnection>(this.CreateConnectionAsync, CloseConnection);
--- a/src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
+++ b/src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.Azure.ServiceBus
 {
     using System;
+    using System.Collections.Generic;
     using System.Text;
 
     /// <summary>
@@ -14,11 +15,19 @@ namespace Microsoft.Azure.ServiceBus
         const char KeyValueSeparator = '=';
         const char KeyValuePairDelimiter = ';';
         static readonly string EndpointScheme = "amqps";
-        static readonly string EndpointFormat = EndpointScheme + "://{0}.servicebus.windows.net";
         static readonly string EndpointConfigName = "Endpoint";
         static readonly string SharedAccessKeyNameConfigName = "SharedAccessKeyName";
         static readonly string SharedAccessKeyConfigName = "SharedAccessKey";
         static readonly string EntityPathConfigName = "EntityPath";
+
+        string entityPath, sasKeyName, sasKey, endpoint;
+
+        /// <summary>
+        /// Instantiates a new <see cref="ServiceBusConnectionStringBuilder"/>
+        /// </summary>
+        public ServiceBusConnectionStringBuilder()
+        {
+        }
 
         /// <summary>
         /// Instatiates a new <see cref="ServiceBusConnectionStringBuilder"/>.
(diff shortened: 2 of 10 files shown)


src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -35,56 +44,93 @@ public ServiceBusConnectionStringBuilder(string connectionString)
...
                 throw Fx.Exception.ArgumentNullOrWhiteSpace(string.IsNullOrWhiteSpace(sharedAccessKeyName) ? nameof(sharedAccessKeyName) : nameof(sharedAccessKey));
             }
-
-            if (namespaceName.Contains("."))
-            {
-                // It appears to be a fully qualified host name, use it.
-                this.Endpoint = new Uri(EndpointScheme + "://" + namespaceName);
-            }
-            else
+            if (!endpoint.Contains("."))
```
vinaysurya: >if (!endpoint.Contains(".")) [](start = 12, length = 28)

Remove this check, the check in the Endpoint setter should take care of this

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -14,12 +15,20 @@ public class ServiceBusConnectionStringBuilder
...
         static readonly string SharedAccessKeyNameConfigName = "SharedAccessKeyName";
         static readonly string SharedAccessKeyConfigName = "SharedAccessKey";
         static readonly string EntityPathConfigName = "EntityPath";
 
+        string entityPath, sasKeyName, sasKey, endpoint;
+
+        /// <summary>
+        /// Instantiates a new <see cref="ServiceBusConnectionStringBuilder"/>
+        /// </summary>
+        public ServiceBusConnectionStringBuilder()
```
vinaysurya: >public [](start = 8, length = 6)

If not using this now, may be make this internal, we could make this public later, when we use it right?

vinaysurya:
:shipit:

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
nemakam: As discussed offline, keeping this as is for now.

SeanFeldman: @nemakam @vinaysurya from a public API point of view, since this is a public class, it's better to seal/internalize members/methods, unless known for sure those need to be public. So 👎 for keeping it as-is (i.e. `public`) if there's no real reason for that.

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -51,7 +51,7 @@ public Task CloseAsync()
 
         protected void InitializeConnection(ServiceBusConnectionStringBuilder builder)
         {
-            this.Endpoint = builder.Endpoint;
+            this.Endpoint = new Uri(builder.Endpoint);
```
nemakam: Sure we could have a custom type for lot of things, but IMHO, that separate type won't have much more functionality than what Uri class provides.

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
nemakam: @SeanFeldman , we have left it public for the sake of flexibility. Right now we are forced to use the constructor which accepts sasKeyName and sasKeyValue (if you don't have the entire connection string). In the near future, we will have to expose other types of authentication including sasTokens. Restricting to use a `builder` unless you have all the necessary variables is no longer a `builder` and strictly becomes a formatter. You lose the flexibility of what a builder is supposed to do.

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
SeanFeldman: @nemakam I made reference to the `ServiceBusConnection.NamespaceName` I've [added](https://github.com/Azure/azure-service-bus-dotnet/pull/193/files#diff-c938c0de58dbc4dacfdb4361f800aceaR106). While `Uri` does it all, it's not SBConnection's responsibility to determine the namespace name I'd think. Not a prio, but something I wanted to raise 🙂

nemakam: @SeanFeldman Thanks for pointing out. That needs to change to show the entire namespace. I remembered that was the reason why I had put the PR on block. 😉

SeanFeldman: Aha 😃 Let's get this one finished and then get back to #193

Issue #189, opened 2017-06-16
MessageReceiver.ReceiveAsync(serverWaitTime) doesn't accept value less than a second


SeanFeldman:
Connect to #183 (see [comment](https://github.com/Azure/azure-service-bus-dotnet/pull/183#discussion_r122316536))

The fix for #154 and #177 reduced `serverWaitTime` from 1 minute to 1 seconds.
Attempts to reduce it to less than a second or setting to `TimeSpan.Zero` resulted in tests failure.
Can `MessageReceiver.ReceiveAsync(serverWaitTime)` accept values less than 1 second?

nemakam:
@SeanFeldman Typo in the heading? 
"value less than a second"

vinaysurya:
Pull Request #199 should resolve this issue

Issue #186, opened 2017-06-14
TaskExtensionHelper.Schedule is used incorrectly


SeanFeldman:
The helper method [`TaskExtensionHelper.Schedule`](https://github.com/Azure/azure-service-bus-dotnet/blob/9ffe9d9ded13f16ba07104f05b69d880e9512dbb/src/Microsoft.Azure.ServiceBus/Primitives/TaskExtensionHelper.cs) introduces an option of safe "fire and forget" execution of async methods.

The way it does it is by using `Task.Run`.

`Task.Run` shouldn't be used for IO-bound operations, but for CPU-bound operations ([see here](https://channel9.msdn.com/Series/Three-Essential-Tips-for-Async/Tip-2-Distinguish-CPU-Bound-work-from-IO-bound-work)).

Rather than wrapping a call with `Task.Run` and doing try catch, task continuation can be used to swallow exception and IO operation is better to be `await`-ed.

PR #199, opened 2017-06-28
Fix on message registration delay issue

vinaysurya:
## Description
Address issue #189 
Do not block waiting for a message/session when Registering a message/session handler.
Also make the exception handling contract a must when registering an on OnMessage/OnSession handler() so that user thinks about exceptions that might happen on the OnMessage Pump right from the getgo. 

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
@@ -44,10 +44,10 @@ namespace Microsoft.Azure.ServiceBus
                 this.registerHandlerOptions.AutoRenewLock;
         }
 
-        async Task RaiseExceptionReceived(Exception e, string action)
+        Task RaiseExceptionReceived(Exception e, string action)
         {
             var eventArgs = new ExceptionReceivedEventArgs(e, action, this.endpoint, this.messageReceiver.Path, this.messageReceiver.ClientId);
-            await this.registerHandlerOptions.RaiseExceptionReceived(eventArgs).ConfigureAwait(false);
+            return this.registerHandlerOptions.RaiseExceptionReceived(eventArgs);
         }
 
         async Task MessagePumpTaskAsync()
--- a/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
@@ -32,15 +32,9 @@ namespace Microsoft.Azure.ServiceBus
             this.maxConcurrentCallsSemaphoreSlim = new SemaphoreSlim(this.registerHandlerOptions.MaxConcurrentCalls);
         }
 
-        public async Task StartPumpAsync()
+        public void StartPump()
         {
-            var initialMessage = await this.messageReceiver.ReceiveAsync(Constants.MessageReceiverStartPumpInitialReceiveTimeout).ConfigureAwait(false);
-            if (initialMessage != null)
-            {
-                MessagingEventSource.Log.MessageReceiverPumpInitialMessageReceived(this.messageReceiver.ClientId, initialMessage);
-            }
-
-            TaskExtensionHelper.Schedule(() => this.MessagePumpTask(initialMessage));
+            TaskExtensionHelper.Schedule(() => this.MessagePumpTaskAsync());
         }
 
         bool ShouldRenewLock()
@@ -50,7 +44,7 @@ namespace Microsoft.Azure.ServiceBus
                 this.registerHandlerOptions.AutoRenewLock;
         }
 
-        async Task MessagePumpTask(Message initialMessage)
+        async Task MessagePumpTaskAsync()
         {
             while (!this.pumpCancellationToken.IsCancellationRequested)
             {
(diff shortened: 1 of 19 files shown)


src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
```diff
@@ -32,15 +32,9 @@ sealed class MessageReceivePump
...
+        public void StartPump()
         {
-            var initialMessage = await this.messageReceiver.ReceiveAsync(Constants.MessageReceiverStartPumpInitialReceiveTimeout).ConfigureAwait(false);
-            if (initialMessage != null)
-            {
-                MessagingEventSource.Log.MessageReceiverPumpInitialMessageReceived(this.messageReceiver.ClientId, initialMessage);
-            }
-
-            TaskExtensionHelper.Schedule(() => this.MessagePumpTask(initialMessage));
+            TaskExtensionHelper.Schedule(() => this.MessagePumpTaskAsync());
```
SeanFeldman: With #186 in place, is this the right thing to do?

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
```diff
@@ -1,32 +1,37 @@
...
+        public string Action { get; private set; }
+
+        /// <summary>The namespace name used when this exception occured.</summary>
+        public string Endpoint { get; private set; }
+
+        /// <summary>The entity path used when this exception occured.</summary>
+        public string EntityPath { get; private set; }
+
+        /// <summary>The Client Id associated with the sender, receiver or session when this exception occured.</summary>
+        public string ClientId { get; private set; }
```
SeanFeldman: Client ID is used for object identification, right?
By reading xmldoc I would not understand why this is exposed and what would I do with it when exception is thrown.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -93,10 +72,10 @@ bool ShouldRenewSessionLock()
                 this.sessionHandlerOptions.AutoRenewLock;
         }
 
-        void RaiseExceptionReceived(Exception e, string action)
+        async Task RaiseExceptionReceived(Exception e, string action)
         {
-            var eventArgs = new ExceptionReceivedEventArgs(e, action, this.endpoint, this.entityPath);
-            this.sessionHandlerOptions.RaiseExceptionReceived(eventArgs);
+            var eventArgs = new ExceptionReceivedEventArgs(e, action, this.endpoint, this.entityPath, this.clientId);
+            await this.sessionHandlerOptions.RaiseExceptionReceived(eventArgs).ConfigureAwait(false);
```
SeanFeldman: It it's invoked and no need in the awaited result, method can drop `async` and return the task

`return this.sessionHandlerOptions.RaiseExceptionReceived(eventArgs)`

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
SeanFeldman: Client ID is used for object identification, right?
By reading xmldoc I would not understand why this is exposed and what would I do with it when exception is thrown.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
SeanFeldman: With #186 in place, is this the right thing to do?

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
vinaysurya: Would rewording to "The Client Id can be used to associate the exception with the queueclient, subscriptionClient, MessageSender or MessageReceiver  that encountered the exception"

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
vinaysurya: I haven't spent time on 186, but since we are not changing that behavior here,  I would prefer to handle that as a separate issue itself.

nemakam: @SeanFeldman IMO, we should tackle this and #186 separately.

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
SeanFeldman: 👍 
Still, in production, where let's say it's logged, how this piece of information can help? I might ask a stupid question, but this is what I'd ask when get this info in a log stream.

src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
SeanFeldman: OK

vinaysurya:
Good point about the sample update. I've created a tracking issue on that 
https://github.com/Azure/azure-service-bus/issues/46

src/Microsoft.Azure.ServiceBus/Primitives/ExceptionReceivedContext.cs
vinaysurya: Basically lets say you create multiple  Queueclients and want to figure out which of the queueclients are throwing that exception for you, then this identifier can be used to figure that out.

SeanFeldman: Wouldn't that only be possible when you have debugger attached?

vinaysurya: If you are using logs, we log clientId's with most events right?

vinaysurya: logs as in EventSource events i mean

SeanFeldman:
BTW, https://github.com/blog/2389-requesting-reviews-from-repository-teams

Issue #109, opened 2017-03-24
Support for safe-size batch


jtaubensee:
## Actual Behavior
1. I have reason to believe that Message.Size will only calculate the size of a stream, and not the headers.

## Expected Behavior
1. Message.Size should include the size of the headers

## Versions
All

SeanFeldman:
👍 
This is an old pain. The current (we should really find a name to call it) client does a poor job, especially when querying the messages that were not sent yet.

jtaubensee:
So, after looking into this one a little bit more, we are wondering where would be the best place to put this functionality. I think rather than being a property on the message, this should be a "utility function" somewhere (perhaps on the `MessageSender`). The reason being is that the size cannot be cached (as the body and properties are mutable), and in order to provide the true message size, we need to convert the message to an AMQP message first. Of course, there is some overhead associated with this, hence the reason to have a method called 'CalculateMessageSize`.

Any thoughts about where this should go? Also, keep in mind that we will throw an exception on the client side before the message is sent if the size is too big. So in most cases, users can just handle the exception on send.

SeanFeldman:
The message size is important to be assessed _prior_ to sending it. Batching allows up to 100 messages or less than the maximum size for the tier used. If there are less than 100 messages, one has to calculate the overall size of the messages in order to make a conscious decision to split the batch into "chunks".

Utility method sounds off to be frank. The message is the one that should know how to calculate its size, not some method on an arbitrary class that might (or might not) be used later. Yes, calculation of the message time is a perf hit, but it's a necessary evil. And when someone opts into it, they probably do that for a good reason.

To sum it up:
- No to a utility method on something like `MessageSender`
- No to an assumption at what point that calculated size will be needed
- Yes for the `Message` to provide this method and proper documentation of what it's doing

If you really oppose the idea of `Message` providing the calculation, you could implement it as an extension method on the root namespace (along with the `Message`).

SeanFeldman:
@Azure/azure-service-bus-write this hasn't see any action. Should it be addressed?

poveilleux:
Any news on that?

SeanFeldman:
Has been backloged @poveilleux 🙁 
What's the use case you're having?

poveilleux:
[...]

PR #200, opened 2017-06-30
API documentation + Several minor API changes

nemakam:
## Description
I've added extra documentation to the client and some examples for the clients.

I've also changed few APIs:
1. `Message.Size` -> Returning body size for now till we have a better solution.
2. Removed `Message.Publisher` as it is an internal property even in the old client and is used for internal purposes.
3. Moved `Message.DeadLetterSource` under `Message.SystemProperties.DeadLetterSource`.
4. Moved public properties of `MessageReceiver.LockedUntilUtc` and `MessageReceiver.SessionId` to under `MessageSession` class as they are specific to sessions.
5. Added plugin related methods to `IClientEntity` 
6. Moved `SubscriptionClient.DefaultRule` to `RuleDescription.DefaultRuleName` as that seems to be a better place
7. Made `IMessageSender` which is the interface for `MessageSender` public.
8. Moved `OperationTimeout` to `IClientEntity` as now every client opens their own connection and we do not allow users to pass in their connections.


src/Microsoft.Azure.ServiceBus/Filters/RuleDescription.cs
```diff
@@ -100,7 +111,7 @@ public string Name
 
             set
             {
-                if (string.IsNullOrWhiteSpace(value))
+                if (String.IsNullOrWhiteSpace(value))
                 {
```
SeanFeldman: Not sure about this one

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -211,9 +200,13 @@ public TimeSpan TimeToLive
 
         // TODO: Calculate the size of the properties and body
         /// <summary>
-        /// Gets the total size of the message in bytes.
+        /// Gets the total size of the message body in bytes.
```
SeanFeldman: Perhaps a remark about actual message size vs this property?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -359,6 +347,23 @@ internal set
                 }
             }
 
+            /// <summary>
+            /// Gets the name of the queue or subscription that this message was enqueued on, before it was deadlettered.
+            /// </summary>
+            public string DeadLetterSource
```
SeanFeldman: An auto property with a non public getter won't work?

nemakam: Didn't get you? Why non-public getter? We want this to have a public get...

src/Microsoft.Azure.ServiceBus/Filters/RuleDescription.cs
nemakam: @SeanFeldman , that must have got changed automatically during resharper refactoring. Will change that back

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: What do you suggest? What remark would do good?
Current thought process - Its better to have a public property which doesn't really do anything useful, whose implementation can later be improved with the fix of #109; than make this internal and then later introduce a new property when we have a fix. 
Hence, this change where I explicitly say for now that the size is just the body size.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: It was in regards to setter, which I thought was public. My bad.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: Rather than having a property called `Size` that represents body size only and later introduce another property for the total size, call it what is 🙂 - ` BodySize` and add `Size` later.

nemakam: Just rewording what I meant - 
Don't want to change the interface/public properties later once we hit GA, and that's why keeping this property public instead of making it internal. 
Now given that we're keeping it public, just making it return BodySize for now instead of 0 till #109 is resolved. Once we work on #109, the same property should return the entire size of message and will also update the summary to mention the same.

SeanFeldman: :+1:

src/Microsoft.Azure.ServiceBus/IMessageSession.cs
```diff
@@ -9,8 +9,21 @@ namespace Microsoft.Azure.ServiceBus
...
     /// <summary>
-    /// Interface used to describe a MessageSession.
+    /// Describes a Session object. IMessageSession can be used to perform operations on sessions.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Service Bus Sessions, also called "Groups" in the AMQP 1.0 protocol, are unbounded sequences of related messages. ServiceBus guarantees ordering of messages in a session.
+    /// </para>
+    /// <para>
+    /// Any sender can "create" a session when submitting messages into a Topic or Queue by setting the <see cref="Message.SessionId"/> property on Message to some
```
vinaysurya: >"create" [](start = 23, length = 8)

nit: remove double quotes

src/Microsoft.Azure.ServiceBus/IMessageSession.cs
```diff
@@ -9,8 +9,21 @@ namespace Microsoft.Azure.ServiceBus
...
-    /// Interface used to describe a MessageSession.
+    /// Describes a Session object. IMessageSession can be used to perform operations on sessions.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Service Bus Sessions, also called "Groups" in the AMQP 1.0 protocol, are unbounded sequences of related messages. ServiceBus guarantees ordering of messages in a session.
+    /// </para>
+    /// <para>
+    /// Any sender can "create" a session when submitting messages into a Topic or Queue by setting the <see cref="Message.SessionId"/> property on Message to some 
+    /// application-defined identifier that is unique to the session. At the AMQP 1.0 protocol level, this value maps to the group-id property.
```
vinaysurya: >identifier that is unique to the session [](start = 28, length = 40)

condense to application defined unique identifier

src/Microsoft.Azure.ServiceBus/IMessageSession.cs
```diff
@@ -19,27 +32,38 @@ public interface IMessageSession : IMessageReceiver
...
         Task SetStateAsync(Stream sessionState);
 
         /// <summary>
-        /// Renews the session lock.
+        /// Renews the lock on the session specified by the <see cref="SessionId"/>. The lock will be renewed based on the setting specified on the entity.
         /// </summary>
-        /// <returns>The asynchronous operation</returns>
+        /// <remarks>
+        /// <para>
+        /// When you accept/receive a session, the session is locked for this client instance by the service for a duration as specified during the Queue/Subscription creation.
```
vinaysurya: >accept/receive [](start = 21, length = 14)

lets just pick one, say accept(since api is accept) and use that

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -30,7 +31,7 @@ public Message()
         /// <summary>
         /// Creates a new message from the specified payload.
         /// </summary>
-        /// <param name="body"></param>
+        /// <param name="body">The payload of the message in byte[]</param>
```
vinaysurya: >byte[] [](start = 61, length = 6)

bytes

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -86,12 +83,10 @@ public string PartitionKey
 
         /// <summary>Gets or sets a partition key value when a transaction is to be used to send messages via a transfer queue.</summary>
```
vinaysurya: >to be [](start = 78, length = 5)

"is used to " is good enough

vinaysurya: > when a transaction is to be used to send [](start = 55, length = 41)

use similar language as above for PartitionKey.like "for sending a transactional message via a transfer queue"

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -135,6 +126,7 @@ public string ReplyToSessionId
         /// <summary>Gets the date and time in UTC at which the message is set to expire.</summary>
         /// <value>The message expiration time in UTC.</value>
         /// <exception cref="System.InvalidOperationException">If the message has not been received. For example if a new message was created but not yet sent and received.</exception>
+        /// <remarks>Unless specifically set for a message, this value is controlled by the default message TTL value set on the entity.</remarks>
```
vinaysurya: >default message TTL value set on the entity [](start = 92, length = 43)

'DefaultMessageTimeToLive'  property set while creating the entity

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -148,8 +140,13 @@ public DateTime ExpiresAtUtc
             }
         }
 
-        /// <summary>Gets or sets the message’s time to live value. This is the duration after which the message expires, starting from when the message is sent to the Service Bus. Messages older than their TimeToLive value will expire and no longer be retained in the message store. Subscribers will be unable to receive expired messages. TimeToLive is the maximum lifetime that a message can be received, but its value cannot exceed the entity specified 
-        /// value on the destination queue or subscription. If a lower TimeToLive value is specified, it will be applied to the individual message. However, a larger value specified on the message will be overridden by the entity’s DefaultMessageTimeToLive value.</summary> 
+        /// <summary>
+        /// Gets or sets the message’s time to live value. This is the duration after which the message expires, starting from when the message is sent to the Service Bus. 
+        /// Messages older than their TimeToLive value will expire and no longer be retained in the message store. Subscribers will be unable to receive expired messages.
```
vinaysurya: >Subscribers [](start = 115, length = 11)

Also queues, so may be word it as Expired messages are not receivable

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -282,18 +275,13 @@ private static void ValidatePartitionKey(string partitionKeyPropertyName, string
         public sealed class SystemPropertiesCollection
         {
             private int deliveryCount;
```
vinaysurya: >private [](start = 12, length = 7)

nit: remove privates here

src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
```diff
@@ -4,12 +4,13 @@
...
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
     using Primitives;
 
     /// <summary>Provides options associated with message pump processing using
-    /// <see cref="QueueClient.RegisterMessageHandler(System.Func{Microsoft.Azure.ServiceBus.Message,System.Threading.CancellationToken,System.Threading.Tasks.Task},MessageHandlerOptions)" /> and
-    /// <see cref="SubscriptionClient.RegisterMessageHandler(System.Func{Microsoft.Azure.ServiceBus.Message,System.Threading.CancellationToken,System.Threading.Tasks.Task},MessageHandlerOptions)" />.</summary>
+    /// <see cref="QueueClient.RegisterMessageHandler(Func{Message,CancellationToken,Task},MessageHandlerOptions)" /> and
```
vinaysurya: >,Can [](start = 66, length = 4)

nit: spaces plz

src/Microsoft.Azure.ServiceBus/MessageSession.cs
```diff
@@ -21,6 +21,24 @@ public MessageSession(string sessionId, DateTime lockedUntilUtc, MessageReceiver
             this.ReceiveMode = innerMessageReceiver.ReceiveMode;
         }
 
+        /// <summary>
+        /// Gets the DateTime that the current receiver is locked until. This is only applicable when Sessions are used.
```
vinaysurya: >current receiver [](start = 39, length = 16)

session

vinaysurya: >This is only applicable when Sessions are used. [](start = 73, length = 47)

this is description on sessions in the MessageSession file, so we can skip this part

src/Microsoft.Azure.ServiceBus/QueueClient.cs
```diff
@@ -31,6 +69,7 @@ public class QueueClient : ClientEntity, IQueueClient
         /// <param name="connectionStringBuilder"><see cref="ServiceBusConnectionStringBuilder"/> having namespace and queue information.</param>
         /// <param name="receiveMode">Mode of receive of messages. Defaults to <see cref="ReceiveMode"/>.PeekLock.</param>
         /// <param name="retryPolicy">Retry policy for queue operations. Defaults to <see cref="RetryPolicy.Default"/></param>
+        /// <remarks>Creates a new connection to the queue, which is opened during the first operation.</remarks>
```
vinaysurya: > first operation. [](start = 86, length = 17)

first send/receive operation

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -35,6 +64,7 @@ public class SubscriptionClient : ClientEntity, ISubscriptionClient
         /// <param name="subscriptionName">Name of the subscription.</param>
         /// <param name="receiveMode">Mode of receive of messages. Defaults to <see cref="ReceiveMode"/>.PeekLock.</param>
         /// <param name="retryPolicy">Retry policy for subscription operations. Defaults to <see cref="RetryPolicy.Default"/></param>
+        /// <remarks>Creates a new connection to the subscription, which is opened during the first operation.</remarks>
```
vinaysurya: >operation [](start = 100, length = 9)

send/receive operation

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -233,74 +281,111 @@ public Task CompleteAsync(string lockToken)
...
 
         /// <summary>
-        /// Asynchronously adds a rule to the current subscription with the specified name and filter expression.
+        /// Adds a rule to the current subscription to filter the messages reaching from topic to the subscription.
         /// </summary>
         /// <param name="ruleName">The name of the rule to add.</param>
         /// <param name="filter">The filter expression against which messages will be matched.</param>
         /// <returns>A task instance that represents the asynchronous add rule operation.</returns>
+        /// <remarks>
+        /// You can add rules to the subscription that will decide filter which messages from the topic should reach the subscription.
```
vinaysurya: >decide [](start = 60, length = 6)

nit: typo

src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
```diff
@@ -8,101 +8,159 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+        /// <returns>Message with the sequence number <paramref name="sequenceNumber"/>. Returns null if no such message is found.</returns>
         Task<Message> ReceiveBySequenceNumberAsync(long sequenceNumber);
 
         /// <summary>
-        /// Receives an <see cref="IList{Message}"/> of messages using the <see cref="MessageReceiver" />.
+        /// Receives a <see cref="IList{Message}"/> messages identified by <paramref name="sequenceNumbers"/>.
         /// </summary>
         /// <param name="sequenceNumbers">An <see cref="IEnumerable{T}"/> containing the sequence numbers to receive.</param>
-        /// <returns>The asynchronous operation.</returns>
+        /// <returns>Messages whoese sequence numbers were requested. Returns null if no messages are found.</returns>
```
vinaysurya: >whoese [](start = 30, length = 6)

nit: spelling

vinaysurya: >whoese sequence numbers were requested [](start = 30, length = 38)

Messages identified by sequence number are returned

src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
```diff
@@ -8,101 +8,159 @@ namespace Microsoft.Azure.ServiceBus.Core
...
+        /// <remarks> Receving less than <paramref name="maxMessageCount"/> messages is not an indication of empty entity.</remarks>
+        Task<IList<Message>> ReceiveAsync(int maxMessageCount, TimeSpan operationTimeout);
 
         /// <summary>
-        /// Receives a message using the <see cref="MessageReceiver" />.
+        /// Receives a particular message identified by <paramref name="sequenceNumber"/>.
         /// </summary>
         /// <param name="sequenceNumber">The sequence number of the message that will be received.</param>
-        /// <returns>The asynchronous operation.</returns>
+        /// <returns>Message with the sequence number <paramref name="sequenceNumber"/>. Returns null if no such message is found.</returns>
```
vinaysurya: >Message with the sequence number [](start = 21, length = 32)

Messages identified by sequence number

vinaysurya:
:clock1:

PR #201, opened 2017-06-30
MessageReceiver should not return expired messages and should retry until it gets a message

vinaysurya:
## Description
This addresses the following tracking bugs:
1) MessageReceiver should not return expired messages 
2) MessageReceiver should retry until remaining time to get a message
3) Renew messages more quickly if the remaining expiry time is less
4) Pass ClientId into linkname for tracking purposes


src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -671,7 +671,7 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
                 IEnumerable<AmqpMessage> amqpMessages = null;
                 IList<Message> brokeredMessages = null;
 
-                while (timeoutHelper.RemainingTime() >= TimeSpan.Zero)
+                while (timeoutHelper.RemainingTime() > TimeSpan.Zero)
```
nemakam: Good catch

PR #207, opened 2017-07-02
Message Extension for interop with older .Net Clients

vinaysurya:
## Description
This PR is related to PR https://github.com/Azure/azure-service-bus-dotnet/issues/114  to provide interop for messages sent through older full DotNet clients. Previous full DotNetClients used a custom DataContractBinarySerializer or XmlObjectSerializer to serialize any objects that were sent to ServiceBus queue/Topic. However, this client always returned the message body as an array of message bytes. So we decided to provided an extension method on the message such that the receive of messages from new client has a way to work with the older messages.


test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterOpTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        void VerifyReturnedObject(Book originalObject, Book returnedObject)
+        {
+            Assert.True(string.Equals(returnedObject.Name, originalObject.Name));
+            Assert.True(originalObject.Id == returnedObject.Id);
+            Assert.True(originalObject.Count == returnedObject.Count);
+        }
+
+        [DataContract(Name = "Library", Namespace = "Books")]
```
SeanFeldman: Is DataContract necessary?
The old client could take a POCO.

src/Microsoft.Azure.ServiceBus/Extensions/DataContractBinarySerializer.cs
```diff
@@ -0,0 +1,125 @@
...
+    using System.Runtime.Serialization;
+    using System.Xml;
+
+    /// <summary>
+    /// This class describes a serializer class used to serialize and deserialize an Object.
+    /// This class is almost identical to DataContractSerializer; only difference is that 
+    /// ReadObject(Stream) and WriteObject(Stream, object) pick Binary Xml Reader/Writer 
+    /// instead of text.
+    /// </summary>
+    public sealed class DataContractBinarySerializer : XmlObjectSerializer
```
SeanFeldman: This is public. I assume this is intentional?

SeanFeldman:
@vinaysurya you might want to resent key for `vinsuaftersendfix` 😉

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Extensions
```
binzywu: hi Sean,

We want to encourage the app to handle the serialization instead of service bus client library. So that people can do necessary optimization and any customization. e.g. nowadays, people might prefer json while performance is not a particular concern. However, we are also skeptically whether we actually should provide some easy way to do so. So this is a starting point to at least support inter-op with old .Net library's data contract serialization.

vinaysurya: @SeanFeldman, yes it was a deliberate choice.

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterOpTests.cs
vinaysurya: Didn't know about POCO.. Corrected, thanks!

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterOpTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            new object[] { TestConstants.NonPartitionedQueueName },
+            new object[] { TestConstants.PartitionedQueueName}
+        };
+
+        [Theory]
+        [MemberData(nameof(TestSerializerPermutations))]
+        [DisplayTestMethodName]
+        void RunSerializerTests(XmlObjectSerializer serializer)
+        {
+            Book book = new Book("contoso", 1, 5);
```
vinaysurya: I prefer writing the full name but this is test class, so not to worried. Will change

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    /// <summary>
+    /// A Message Extension Class that provides extension methods to deserialize 
+    /// the body of a message that was serialized and sent to Service Bus Queue/Topic
+    /// using the full .Net Client library which uses a DataContractSerializer.
+    /// </summary>
+    public static class MessageInterOpExtensions
+    {
+        /// <summary>
+        /// Deserializes the body of a message that was serialized using DataContractBinarySerializer
```
vinaysurya: Changed

vinaysurya: Added more details and remarks

src/Microsoft.Azure.ServiceBus/Extensions/DataContractBinarySerializer.cs
vinaysurya: Corrected

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterOpTests.cs
```diff
@@ -0,0 +1,105 @@
...
+                serializer.WriteObject(stream, bookObject);
+                stream.Flush();
+                stream.Position = 0;
+                payload = stream.ToArray();
+            };
+
+            return new Message(payload);
+        }
+
+        void VerifyReturnedObject(Book originalObject, Book returnedObject)
```
vinaysurya: good idea, done!

vinaysurya:
@SeanFeldman, @binzywu, let me know if you have any further comments on the PR

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
SeanFeldman: In that case, perhaps namespace should be something like `Microsoft.Azure.ServiceBus.Interop`?  Not in the main space, but indicates what it stands for.

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterOpTests.cs
```diff
@@ -0,0 +1,105 @@
...
+            Book returnedBookObject = message.GetBody<Book>(serializer);
+            VerifyReturnedObject(book, returnedBookObject);
+        }
+
+        [Theory]
+        [MemberData(nameof(TestEnd2EndEntityPermutations))]
+        [DisplayTestMethodName]
+        async Task RunEnd2EndSerializerTests(string queueName)
+        {
+            MessageSender messageSender = new MessageSender(TestUtility.NamespaceConnectionString, queueName);
```
SeanFeldman: Exactly the thought I had @binzywu , just wasn't sure if there's any objection to referencing the old client under the project. Also, the test would only execute under full FW, not the core client.

test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
```diff
@@ -423,6 +423,15 @@ namespace Microsoft.Azure.ServiceBus.Core
         public virtual System.Threading.Tasks.Task<Microsoft.Azure.ServiceBus.Message> BeforeMessageSend(Microsoft.Azure.ServiceBus.Message message) { }
     }
 }
+namespace Microsoft.Azure.ServiceBus.Extensions
+{
+    
+    public class static MessageInteropExtensions
+    {
+        public static T GetBody<T>(this Microsoft.Azure.ServiceBus.Message message) { }
+        public static T GetBody<T>(this Microsoft.Azure.ServiceBus.Message message, System.Runtime.Serialization.XmlObjectSerializer serializer) { }
```
SeanFeldman: Once comment from @binzywu is addressed, this will need to be re-approved to remove this overload.

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        public static T GetBody<T>(this Message message)
+        {
+            return GetBody<T>(message, new DataContractBinarySerializer(typeof(T)));
+        }
+
+        /// <summary>
+        /// Deserializes the body of a message that was serialized using XmlObjectSerializer
+        /// </summary>
+        public static T GetBody<T>(this Message message, XmlObjectSerializer serializer)
```
SeanFeldman: ==>

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
vinaysurya: I deliberately chose extensions for the Uber class thinking if we add any other extensions that are not related to Interop and not related to message, we might be able to put them under the same extensions namespace

SeanFeldman: >  we might be able to put them under the same extensions namespace

"What if" case? 🙂
Reason I've suggested this change to the namespace is because would rather see partitioning by major role (Interop, connectivity, etc) rather than grouping by implementation (extension methods, singletons, etc).

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    ///     await messageSender.SendAsync(message1);
+    ///     await messageSender.SendAsync(message2);
+    ///     await messageSender.SendAsync(message3);
+    ///     
+    /// Then retreive the original objects using this client library as follows:
+    /// (Serializer passed for GetBody() call can be null if message was sent using WindowsAzure.Messaging client
+    /// library and TransportType was AMQP.)
+    /// 
+    ///     var message1 = await messageReceiver.ReceiveAsync();
+    ///     var serializer1 = new DataContractBinarySerializer(typeof(string));
```
SeanFeldman: Question: if in all usages `DataContractBinarySerializer` is required, why not to simplify the extension method and make it closer to what the old API was? I.e. 
`message.GetBody<type>()` which internally news up `DataContractBinarySerializer`. At the same time providing an overload that takes an alternative serializer?

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterop/MessageInteropEnd2EndTests.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            string queueName = TestConstants.NonPartitionedQueueName;
+
+            // Override and Create a new ConnectionString with SbmpConnection Endpoint scheme
+            string[] temp = TestUtility.NamespaceConnectionString.Split(':');
+            string sbConnectionString = "Endpoint=sb:" + temp[1];
+
+            if(transportType == TransportType.Amqp)
+            {
+                sbConnectionString += ';' + nameof(TransportType) + '=' + TransportType.Amqp.ToString();
```
SeanFeldman: > ';' + nameof(TransportType) + '=' + TransportType.Amqp.ToString();

could be passed in as a parameter via permutations parameters, just like `transportType` is.

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterop/MessageInteropEnd2EndTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            // Create a full framework MessageSender
+            MessagingFactory messagingFactory = MessagingFactory.CreateFromConnectionString(sbConnectionString);
+            MessageSender fullFrameWorkClientSender = messagingFactory.CreateMessageSender(queueName);
+
+            // Create a .NetCore MessageReceiver
+            Core.MessageReceiver dotNetStandardMessageReceiver = new Core.MessageReceiver(TestUtility.NamespaceConnectionString, queueName, ServiceBus.ReceiveMode.ReceiveAndDelete);
```
SeanFeldman: .NET Core and .NET Standard mixed in. Oh boy, ubiquitous it is not 🙂

test/Microsoft.Azure.ServiceBus.UnitTests/MessageInterop/MessageInteropEnd2EndTests.cs
```diff
@@ -0,0 +1,109 @@
...
+                await fullFrameWorkClientSender.SendAsync(message3);
+
+                // Stream Object
+                string message4Body = "contosoStreamObject";
+                var message4 = new BrokeredMessage(new MemoryStream(Encoding.UTF8.GetBytes(message4Body)));
+                await fullFrameWorkClientSender.SendAsync(message4);
+
+                // Now Receive using this clients MessageReceiver
+                // Plain string
+                var returnedMessage = await dotNetStandardMessageReceiver.ReceiveAsync();
```
SeanFeldman: Should receiving be done by sequence number? Otherwise test implies order. highly unlikely that messages will come out of order, but still. Alternatively, send/receive pairs? Also will allow to keep the mental scope smaller.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
```diff
@@ -160,33 +160,20 @@ public static SBMessage AmqpMessageToSBMessage(AmqpMessage amqpMessage)
...
-                    else
-                    {
-                        byteArray = new byte[arraySegmentValue.Count];
-                        Array.ConstrainedCopy(arraySegmentValue.Array, arraySegmentValue.Offset, byteArray, 0, arraySegmentValue.Count);
-                    }
-                    
-                    sbMessage = new SBMessage(byteArray);
+                    sbMessage.SystemProperties.BodyObject = amqpMessage.ValueBody.Value;
                 }
                 else
```
nemakam: nit: maybe invert the if - else and remove the `!` from the if condition

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
binzywu: This is a good question. We shall discuss further. For short term, we don't want to limit or provide serialization. The code here is only to support reading messages that were sent from old library. However, future is open to discuss.

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    ///     var message4 = await messageReceiver.ReceiveAsync();
+    ///     string returned = Encoding.UTF8.GetString(message4.Body); // Since message was sent as Stream, no deserialization required here.
+    /// 
+    /// </remarks>
+    public static class MessageInteropExtensions
+    {
+        /// <summary>
+        /// Deserializes the body of a message that was serialized using XmlObjectSerializer
+        /// </summary>
+        public static T GetBody<T>(this Message message, XmlObjectSerializer serializer)
```
binzywu: If we are validating the parameters, better to check serializer too.

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            if(message.SystemProperties.BodyObject != null)
+            {
+                return (T)message.SystemProperties.BodyObject;
+            }
+
+            if(message.Body == null || message.Body.Length == 0)
```
binzywu: as discussed, old client may allow sending null body

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
nemakam: +1 to this. We could have a static serializer as a default serializer so that it becomes easier for majority of the customers.

---
In reply to: [126552059](https://github.com/Azure/azure-service-bus-dotnet/pull/207#discussion_r126552059) [](ancestors = 126552059)

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new InvalidOperationException(Resources.MessageBodyIsNullOrEmpty);
+            }
+
+            using (MemoryStream stream = new MemoryStream(256))
```
nemakam: >256 [](start = 58, length = 3)

instead of 256, use message.Body.Length

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
nemakam: could you update the documentation as well after this

---
In reply to: [126563860](https://github.com/Azure/azure-service-bus-dotnet/pull/207#discussion_r126563860) [](ancestors = 126563860,126552059)

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    /// Scenarios to use the GetBody Extension method:
+    /// ----------------------------------------------
+    /// If message was constructed using the WindowsAzure.Messaging client library as follows:
+    ///     var message1 = new BrokeredMessage("contoso"); // Sending a plain string
+    ///     var message2 = new BrokeredMessage(sampleObject); // Sending an actual customer object
+    ///     var message3 = new BrokeredMessage(Encoding.UTF8.GetBytes("contoso")); // Sending a UTF8 encoded byte array object
+    ///     
+    ///     await messageSender.SendAsync(message1);
+    ///     await messageSender.SendAsync(message2);
+    ///     await messageSender.SendAsync(message3);
```
nemakam: wrap this under <code></code> so that formatting remains proper

src/Microsoft.Azure.ServiceBus/Extensions/MessageInterOpExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    ///     var returnedData1 = message1.GetBody&lt;string&gt;(serializer1);
+    ///     
+    ///     var message2 = await messageReceiver.ReceiveAsync();
+    ///     var serializer2 = new DataContractBinarySerializer(typeof(SampleObject));
+    ///     var returnedData2 = message1.GetBody&lt;SampleObject&gt;(serializer2);
+    ///     
+    ///     var message3 = await messageReceiver.ReceiveAsync();
+    ///     var serializer3 = new DataContractBinarySerializer(typeof(byte[]));
+    ///     var returnedData3Bytes = message1.GetBody&lt;byte[]&gt;(serializer3);
+    ///     Console.WriteLine($"Message3 String: {Encoding.UTF8.GetString(returnedData3Bytes)}");
```
nemakam: wrap under <code>

test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
```diff
@@ -480,9 +480,14 @@ namespace Microsoft.Azure.ServiceBus.Filters
...
     
+    public class static DataContractBinarySerializer<T>
+    
+    {
+        public static System.Runtime.Serialization.XmlObjectSerializer Instance;
+    }
     public class static MessageInteropExtensions
     {
-        public static T GetBody<T>(this Microsoft.Azure.ServiceBus.Message message, System.Runtime.Serialization.XmlObjectSerializer serializer) { }
+        public static T GetBody<T>(this Microsoft.Azure.ServiceBus.Message message, System.Runtime.Serialization.XmlObjectSerializer serializer = null) { }
```
SeanFeldman: ❤️

PR #210, opened 2017-07-06
Moved public properties of SessionId and LockedUntilUtc

nemakam:
## Description
Moved public properties of SessionId and LockedUntilUtc from MessageReceiver to MessageSession.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessageSession.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageSession.cs
@@ -20,13 +20,31 @@ namespace Microsoft.Azure.ServiceBus
             ServiceBusConnection serviceBusConnection,
             ICbsTokenProvider cbsTokenProvider,
             RetryPolicy retryPolicy,
-            int prefetchCount = 0,
+            int prefetchCount = Constants.DefaultClientPrefetchCount,
             string sessionId = null,
             bool isSessionReceiver = false)
             : base(entityPath, entityType, receiveMode, serviceBusConnection, cbsTokenProvider, retryPolicy, prefetchCount, sessionId, isSessionReceiver)
         {
         }
-        
+
+        /// <summary>
+        /// Gets the DateTime that the current receiver is locked until. This is only applicable when Sessions are used.
+        /// </summary>
+        public DateTime LockedUntilUtc
+        {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.ServiceBus/MessageSession.cs
```diff
@@ -20,13 +20,31 @@ internal class MessageSession : MessageReceiver, IMessageSession
...
         {
         }
-        
+
+        /// <summary>
+        /// Gets the time that the session identified by <see cref="SessionId"/> is locked until for this client.
+        /// </summary>
+        public DateTime LockedUntilUtc
+        {
+            get => LockedUntilUtcInternal;
```
vinaysurya: >LockedUntilUtcInternal [](start = 19, length = 22)

nit: this.

src/Microsoft.Azure.ServiceBus/MessageSession.cs
```diff
@@ -20,13 +20,31 @@ internal class MessageSession : MessageReceiver, IMessageSession
...
+            internal set => LockedUntilUtcInternal = value;
+        }
+
+        /// <summary>
+        /// Gets the SessionId.
+        /// </summary>
+        public string SessionId
+        {
+            get => SessionIdInternal;
```
vinaysurya: >SessionIdInterna [](start = 19, length = 16)

nit: this.

vinaysurya:
:shipit:

src/Microsoft.Azure.ServiceBus/MessageSession.cs
SeanFeldman: @vinaysurya is `this` truly needed?

vinaysurya: Personally I like it, because it makes it easier to determine whether its an instance object or static object, without having to go look at the original code place. and the fact that we use this pattern in our service code as well makes it consistent while reading code.. 

So many a times we use this pattern where the constructor and the object use the same name, In those case we anyway end up using it. So if we have to use it in some places, then might as well use it everywhere. 

Ultimately it is a personal preference, and I have not found issues with it and have found it easier to read code, etc ..So I like it.

public Message(byte[] body)
        {
            this.body = body;
         }

PR #211, opened 2017-07-06
Synchronizing the client naming convention across different clients.

nemakam:
## Description
Today different clients initialize with different naming conventions. Synchronizing the client naming convention across different clients. 


src/Microsoft.Azure.ServiceBus/Primitives/ClientEntity.cs
```diff
@@ -107,5 +102,15 @@ protected static long GetNextId()
         {
             return Interlocked.Increment(ref nextId);
         }
+
+        /// <summary>
+        /// Generates a new client id that can be used to identify a specific client in logs and error messages.
+        /// </summary>
+        /// <param name="clientTypeName">The type of the client.</param>
+        /// <param name="postfix">Information that can be appended by the client.</param>
```
SeanFeldman: > Information that can be appended

If it's optional, consider making `postfix` optional

PR #212, opened 2017-07-06
Extending the plugins to messagesessions

nemakam:
## Description
Extending the plugins to MessageSession.
Ensuring the registered plugins gets passed on to the sessions that were accepted.

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/PluginTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/PluginTests.cs
@@ -98,9 +98,12 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                 };
                 await messageSender.SendAsync(sendMessage);
 
-                var receivedMessage = await messageReceiver.ReceiveAsync(1, TimeSpan.FromMinutes(1));
+                // Ensure the plugin is called.
+                Assert.True(sendReceivePlugin.MessageBodies.ContainsKey(sendMessage.MessageId));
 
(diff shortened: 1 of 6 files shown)


test/Microsoft.Azure.ServiceBus.UnitTests/PluginTests.cs
```diff
@@ -152,6 +155,56 @@ public class PluginTests
...
+                        Assert.Equal(sendMessage.SessionId, session.SessionId);
+                        Assert.True(session.RegisteredPlugins.Contains(sendReceivePlugin));
+                        Assert.Equal(sendMessage.Body, message.Body);
+
+                        messageReceived = true;
+                        return Task.CompletedTask;
+                    },
+                    exceptionArgs => Task.CompletedTask);
+
+                for (int i = 0; i < 20; i++)
```
vinaysurya: >(int i = 0; i < 20; i++) [](start = 20, length = 24)

that is quite some retry to receive a session, is it that bad :-)

vinaysurya:
:shipit:

test/Microsoft.Azure.ServiceBus.UnitTests/PluginTests.cs
nemakam: Its not exactly a retry. Wanted to wait for around 30-40 seconds before I throw failure. So instead of 10 times delay of 4 seconds, 20 times delay of 2 seconds seems effective. 😄

PR #216, opened 2017-07-10
Adding additional logging

nemakam:
## Description
Adding additional logging to capture the state during an exception

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
@@ -1180,5 +1180,20 @@ namespace Microsoft.Azure.ServiceBus
         {
             WriteEvent(99, exception);
         }
+
+        [NonEvent]
+        public void LinkStateLost(string clientId, string receiveLinkName, AmqpObjectState receiveLinkState, bool isSessionReceiver, Exception exception)
+        {
+            if (this.IsEnabled())
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
```diff
@@ -1150,6 +1150,21 @@ void ExceptionReceivedHandlerThrewException(string exception)
             WriteEvent(99, exception);
         }
 
+        [NonEvent]
+        public void LinkStateLost(string clientId, string receiveLinkName, AmqpObjectState receiveLinkState, bool isSessionReceiver, Exception exception)
+        {
+            if (this.IsEnabled())
+            {
+                this.LinkStateLost(clientId, receiveLinkName, receiveLinkState.ToString(), isSessionReceiver, exception.ToString());
```
vinaysurya: >exception.ToString() [](start = 110, length = 20)

nit: considering this is OperationCancelled exception. we will probably not have any more details in the exception trace. just logging exception.message will be sufficient here?

vinaysurya:
:shipit:

src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
nemakam: Maybe. But if you are just looking at the logs, having the stack trace around could help in faster debugging. Extra logging might not a big issue anyway in the case of exceptions. Leaving it as-is

PR #226, opened 2017-07-19
Handle the scenario where session link is closed from the broker due to Session timing out

vinaysurya:
## Description
Handle the scenario where session link is closed from the broker due to Session timing out. The client should detect that scenario and throw SessionLockLostExceptions for any further operations on the session.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -5,7 +5,6 @@ namespace Microsoft.Azure.ServiceBus.Core
 {
     using System;
     using System.Collections.Generic;
-    using System.Diagnostics;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -709,8 +708,6 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
 
             receivingAmqpLink.Closed += this.OnSessionReceiverLinkClosed;
-            Debug.WriteLine($"{DateTime.Now}: ReceivingLink: {receivingAmqpLink}");
-
             this.SessionIdInternal = tempSessionId;
             long lockedUntilUtcTicks;
             this.LockedUntilUtcInternal = receivingAmqpLink.Settings.Properties.TryGetValue(AmqpClientConstants.LockedUntilUtc, out lockedUntilUtcTicks) ? new DateTime(lockedUntilUtcTicks, DateTimeKind.Utc) : DateTime.MinValue;
@@ -768,10 +765,8 @@ namespace Microsoft.Azure.ServiceBus.Core
                 TimeoutHelper timeoutHelper = new TimeoutHelper(serverWaitTime, true);              
                 if(!this.ReceiveLinkManager.TryGetOpenedObject(out receiveLink))
                 {
-                    Debug.WriteLine($"{DateTime.Now}: No OpenedObject found: Calling GetOrCreateAsync()");
-                    Console.WriteLine($"{DateTime.Now}: No OpenedObject found: Calling GetOrCreateAsync()");
                     receiveLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
-                }              
+                }
 
                 IEnumerable<AmqpMessage> amqpMessages = null;
                 IList<Message> brokeredMessages = null;
@@ -1127,8 +1122,6 @@ namespace Microsoft.Azure.ServiceBus.Core
         async Task<ReceivingAmqpLink> CreateLinkAsync(TimeSpan timeout)
         {
             FilterSet filterMap = null;
-            Debug.WriteLine($"{DateTime.Now}: CreateLinkAsync() called");
-            Console.WriteLine($"{DateTime.Now}: CreateLinkAsync() called");
             MessagingEventSource.Log.AmqpReceiveLinkCreateStart(this.ClientId, false, this.EntityType, this.Path);
 
             if (this.isSessionReceiver)
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -1030,6 +1030,11 @@ namespace Microsoft.Azure.ServiceBus.Core
 
         async Task DisposeMessagesAsync(IEnumerable<Guid> lockTokens, Outcome outcome)
         {
+            if(this.isSessionReceiver)
+            {
+                this.ThrowIfSessionLockLost();
+            }
+
             TimeoutHelper timeoutHelper = new TimeoutHelper(this.OperationTimeout, true);
             IList<ArraySegment<byte>> deliveryTags = this.ConvertLockTokensToDeliveryTags(lockTokens);
 
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -735,6 +735,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             RequestResponseAmqpLink requestResponseAmqpLink = null;
             if (!this.RequestResponseLinkManager.TryGetOpenedObject(out requestResponseAmqpLink))
             {
+                MessagingEventSource.Log.CreatingNewLink(this.ClientId, this.isSessionReceiver, this.SessionIdInternal, true, this.LinkError);
                 requestResponseAmqpLink = await this.RequestResponseLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false); 
             }
 
@@ -765,6 +766,7 @@ namespace Microsoft.Azure.ServiceBus.Core
                 TimeoutHelper timeoutHelper = new TimeoutHelper(serverWaitTime, true);              
                 if(!this.ReceiveLinkManager.TryGetOpenedObject(out receiveLink))
                 {
+                    MessagingEventSource.Log.CreatingNewLink(this.ClientId, this.isSessionReceiver, this.SessionIdInternal, false, this.LinkError);
                     receiveLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
                 }
 
@@ -1043,6 +1045,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             {
                 if (!this.ReceiveLinkManager.TryGetOpenedObject(out receiveLink))
                 {
+                    MessagingEventSource.Log.CreatingNewLink(this.ClientId, this.isSessionReceiver, this.SessionIdInternal, false, this.LinkError);
                     receiveLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false); 
                 }
 
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -5,6 +5,7 @@ namespace Microsoft.Azure.ServiceBus.Core
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -212,6 +213,8 @@ namespace Microsoft.Azure.ServiceBus.Core
         internal string SessionIdInternal { get; set; }
 
         internal MessagingEntityType? EntityType { get; private set; }
+
+        Error LinkError { get; set; }
 
         ServiceBusConnection ServiceBusConnection { get; }
 
@@ -690,6 +693,7 @@ namespace Microsoft.Azure.ServiceBus.Core
         {
             TimeoutHelper timeoutHelper = new TimeoutHelper(serverWaitTime, true);
             ReceivingAmqpLink receivingAmqpLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+            
             Source source = (Source)receivingAmqpLink.Settings.Source;
             string tempSessionId;
             if (!source.FilterSet.TryGetValue(AmqpClientConstants.SessionFilterName, out tempSessionId))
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -704,16 +707,36 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
                 throw new ServiceBusException(true, Resources.AmqpFieldSessionId);
             }
 
+            receivingAmqpLink.Closed += this.OnSessionReceiverLinkClosed;
             this.SessionIdInternal = tempSessionId;
             long lockedUntilUtcTicks;
             this.LockedUntilUtcInternal = receivingAmqpLink.Settings.Properties.TryGetValue(AmqpClientConstants.LockedUntilUtc, out lockedUntilUtcTicks) ? new DateTime(lockedUntilUtcTicks, DateTimeKind.Utc) : DateTime.MinValue;
         }
 
+        void OnSessionReceiverLinkClosed(object sender, EventArgs e)
```
nemakam: >OnSessionReceiverLinkClosed [](start = 13, length = 27)

Instead cant we just call `this.CloseSession` and `closeAsync` here? Lets close the link completely, and then each operation should check `clientEntity.IsClosedOrClosing` to take actions .

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1174,6 +1215,15 @@ void ThrowIfNotPeekLockMode()
             }
         }
 
+        void ThrowIfSessionLockLost()
+        {
+            if (this.LinkError != null &&
+                this.LinkError.Condition.Equals(AmqpClientConstants.SessionLockLostError))
+            {
```
nemakam: what if the condition is something else? We still have got a `link.Closed` event. Should this be `if(this.LinkError != null)`?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: 1) I still want the user contract to be 
User attempts operations on a Session.
If Session has expired, User Gets SessionLockLostException.
User already has code to handle SessionLockLost in other scenarios, to Close the Session and Move on.
This way we wont do anything under the covers for the user. and he suddenly discovers the Session is gone.
2) I want to continue throwing SessionLockLostException when User attempts operations on the session. Checking Closed only would mean when operations are attempted on the session, a) I would either not throw SessionLockLostException exception or b) Check that is closing and also check LinkError and throw SessionLockLostException anyway. This looked better.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: I was thinking about that too, but at this point I'm not sure what the other something else is. So until we see issues and fully understand why we need to handle that scenario, I wanted to keep it defined to the scenario we know of.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1174,6 +1218,15 @@ void ThrowIfNotPeekLockMode()
...
```
nemakam: could you add a log over here as well for cases when `LinkError.Condition != SessionLockLost` which captures the actual error state so that we can handle it later

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
```diff
@@ -72,58 +72,55 @@ public class ExpectedMessagingExceptionTests
...
-
-                messageSession = await queueClient.AcceptMessageSessionAsync(sessionId);
-            Assert.NotNull(messageSession);
-
-                message = await messageSession.ReceiveAsync();
-                TestUtility.Log($"Received Message: SessionId: {messageSession.SessionId}");
-
-                await message.CompleteAsync();
+                TestUtility.Log($"Waiting for session lock to time out...");
+                await Task.Delay(sessionReceiver.LockedUntilUtc - DateTime.UtcNow);
```
nemakam: >await Task.Delay(sessionReceiver.LockedUntilUtc - DateTime.UtcNow); [](start = 16, length = 67)

Maybe add a couple of seconds buffer here to avoid flakiness?

nemakam:
:shipit:

src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
```diff
@@ -1173,5 +1174,21 @@ public void UpdateClientId(string oldClientId, string newClientId)
...
+        public void CreatingNewLink(string clientId, bool isSessionReceiver, string sessionId, bool isRequestResponseLink, Error linkError)
+        {
+            if (this.IsEnabled())
+            {
+                string linkErrorString = linkError != null ? linkError.Condition.Value : string.Empty;
+                this.CreatingNewLink(clientId, isSessionReceiver, sessionId ?? string.Empty, isRequestResponseLink, linkErrorString);
+            }
+        }
+
+        [Event(102, Level = EventLevel.Error, Message = "Creating/Recreating New Link. ClientId: {0}, IsSessionReceiver: {1}, SessionId: {2}, IsRequestResponseLink: {3}, LinkError: {4}.")]
```
nemakam: Change this to 105. Get rules is using 102-104

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: Not needed right? Anytime a link is closed and we are recreating a link, we will have the new trace that is describing the link error that was encountered on close anyway.

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
vinaysurya: Had added it initially and removed it purposely so we can see the failure, know why it happenned and then may be add the buffer. That is why Im also tracing the SessionLockedUntilUtc so we can see what is the issue, is it like a timelag or something else.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -704,16 +707,38 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
...
             this.LockedUntilUtcInternal = receivingAmqpLink.Settings.Properties.TryGetValue(AmqpClientConstants.LockedUntilUtc, out lockedUntilUtcTicks) ? new DateTime(lockedUntilUtcTicks, DateTimeKind.Utc) : DateTime.MinValue;
         }
 
+        void OnSessionReceiverLinkClosed(object sender, EventArgs e)
+        {
+            ReceivingAmqpLink amqpLink = (ReceivingAmqpLink)sender;
+            if (amqpLink != null)
+            {
+                AmqpException amqpException = amqpLink.TerminalException as AmqpException;
+                this.LinkError = amqpException != null ? amqpException.Error : null;
```
SeanFeldman: this.LinkError = amqpException?.Error ?? null;

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -704,16 +707,38 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
...
-            RequestResponseAmqpLink requestResponseAmqpLink = await this.RequestResponseLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+            RequestResponseAmqpLink requestResponseAmqpLink = null;
+            if (!this.RequestResponseLinkManager.TryGetOpenedObject(out requestResponseAmqpLink))
+            {
+                MessagingEventSource.Log.CreatingNewLink(this.ClientId, this.isSessionReceiver, this.SessionIdInternal, true, this.LinkError);
+                requestResponseAmqpLink = await this.RequestResponseLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false); 
+            }
 
             AmqpMessage responseAmqpMessage = await Task.Factory.FromAsync(
                 (c, s) => requestResponseAmqpLink.BeginRequest(amqpMessage, timeoutHelper.RemainingTime(), c, s),
```
SeanFeldman: .ConfigureAwait(false);

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -731,10 +756,20 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
         protected virtual async Task<IList<Message>> OnReceiveAsync(int maxMessageCount, TimeSpan serverWaitTime)
         {
             ReceivingAmqpLink receiveLink = null;
+
+            if (this.isSessionReceiver)
```
SeanFeldman: I see this block is repeated multiple times.
Would it help if `ThrowIfSessionLockLost()` would be invoked and condition would be within the method?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -289,9 +289,14 @@ public Task SendAsync(Message message)
 
         internal async Task<AmqpResponseMessage> ExecuteRequestResponseAsync(AmqpRequestMessage amqpRequestMessage)
         {
+            RequestResponseAmqpLink requestResponseAmqpLink = null;
             AmqpMessage amqpMessage = amqpRequestMessage.AmqpMessage;
             TimeoutHelper timeoutHelper = new TimeoutHelper(this.OperationTimeout, true);
-            RequestResponseAmqpLink requestResponseAmqpLink = await this.RequestResponseLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+
+            if (!this.RequestResponseLinkManager.TryGetOpenedObject(out requestResponseAmqpLink))
```
SeanFeldman: if (!this.RequestResponseLinkManager.TryGetOpenedObject(out **var** requestResponseAmqpLink))
to remove unnecessary `RequestResponseAmqpLink requestResponseAmqpLink = null;`

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -704,16 +707,38 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
...
             AmqpMessage amqpMessage = amqpRequestMessage.AmqpMessage;
+            if (this.isSessionReceiver)
+            {
+                this.ThrowIfSessionLockLost();
+            }
+
             TimeoutHelper timeoutHelper = new TimeoutHelper(this.OperationTimeout, true);
-            RequestResponseAmqpLink requestResponseAmqpLink = await this.RequestResponseLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+            RequestResponseAmqpLink requestResponseAmqpLink = null;
+            if (!this.RequestResponseLinkManager.TryGetOpenedObject(out requestResponseAmqpLink))
```
SeanFeldman: if (!this.RequestResponseLinkManager.TryGetOpenedObject(out **var** requestResponseAmqpLink))
to remove unnecessary `RequestResponseAmqpLink requestResponseAmqpLink = null;`

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -731,10 +756,20 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
...
+            {
+                this.ThrowIfSessionLockLost();
+            }
+
             try
             {
-                TimeoutHelper timeoutHelper = new TimeoutHelper(serverWaitTime, true);
-                receiveLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+                TimeoutHelper timeoutHelper = new TimeoutHelper(serverWaitTime, true);              
+                if(!this.ReceiveLinkManager.TryGetOpenedObject(out receiveLink))
```
SeanFeldman: if(!this.ReceiveLinkManager.TryGetOpenedObject(out **var** receiveLink))

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -998,13 +1033,23 @@ public void RegisterMessageHandler(Func<Message, CancellationToken, Task> handle
...
+            }
+
             TimeoutHelper timeoutHelper = new TimeoutHelper(this.OperationTimeout, true);
             IList<ArraySegment<byte>> deliveryTags = this.ConvertLockTokensToDeliveryTags(lockTokens);
 
             ReceivingAmqpLink receiveLink = null;
             try
             {
-                receiveLink = await this.ReceiveLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+                if (!this.ReceiveLinkManager.TryGetOpenedObject(out receiveLink))
```
SeanFeldman: if (!this.ReceiveLinkManager.TryGetOpenedObject(out **var** receiveLink))

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -310,7 +315,10 @@ public Task SendAsync(Message message)
                 SendingAmqpLink amqpLink = null;
                 try
                 {
-                    amqpLink = await this.SendLinkManager.GetOrCreateAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
+                    if (!this.SendLinkManager.TryGetOpenedObject(out amqpLink))
```
SeanFeldman: if (!this.SendLinkManager.TryGetOpenedObject(out **var** amqpLink))

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
vinaysurya: I thought of it initially that way. But then it would be confusing for someone reading MessageReceiver code to say why are we checking for SessionReceiver every time, so I left out the check for SessionReceiver here, so in one glance, people can know whether they should be interested in that or not..

vinaysurya:
Thanks for the comments @SeanFeldman! I will handle the other comments next time I make a checkin..

PR #230, opened 2017-07-24
Fix Azure Resource Manager template link

kierenj:
## Description
In README.md, the link text to download the Azure Resource Manager template for unit tests is /build/azuredeploy.json, but the link itself is /templates/azuredeploy.json.

This minor change brings the link in line, fixing the broken link.


vinaysurya:
Thanks for fixing this @kierenj!

PR #233, opened 2017-07-26
Update MessagingEventSource.cs

jjadeja:
Event 47 was called with 2 argument(s), but it is defined with 3 parameter(s).

## Description


vinaysurya:
Thanks for the fix @jjadeja!

PR #236, opened 2017-07-27
Exposing OperationTimeout property's setter on all clients

nemakam:
## Description
Exposing OperationTimeout property's setter on all clients

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -202,6 +202,14 @@ namespace Microsoft.Azure.ServiceBus.Core
         public virtual string Path { get; private set; }
 
         /// <summary>
+        /// Duration after which individual operations will timeout.
+        /// </summary>
+        public override TimeSpan OperationTimeout {
+            get => this.ServiceBusConnection.OperationTimeout;
+            set => this.ServiceBusConnection.OperationTimeout = value;
+        }
+
+        /// <summary>
         /// Gets the DateTime that the current receiver is locked until. This is only applicable when Sessions are used.
         /// </summary>
         internal DateTime LockedUntilUtcInternal { get; set; }
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -111,6 +111,15 @@ namespace Microsoft.Azure.ServiceBus.Core
         /// </summary>
         public virtual string Path { get; private set; }
 
+        /// <summary>
+        /// Duration after which individual operations will timeout.
+        /// </summary>
+        public override TimeSpan OperationTimeout
+        {
+            get => this.ServiceBusConnection.OperationTimeout;
+            set => this.ServiceBusConnection.OperationTimeout = value;
+        }
+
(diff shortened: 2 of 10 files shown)


src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -202,6 +202,14 @@ internal set
         public virtual string Path { get; private set; }
 
         /// <summary>
+        /// Duration after which individual operations will timeout.
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -111,6 +111,15 @@ public class MessageSender : ClientEntity, IMessageSender
         /// </summary>
         public virtual string Path { get; private set; }
 
+        /// <summary>
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/QueueClient.cs
```diff
@@ -126,6 +126,15 @@ public QueueClient(string connectionString, string entityPath, ReceiveMode recei
         public ReceiveMode ReceiveMode { get; }
 
         /// <summary>
+        /// Duration after which individual operations will timeout.
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/SessionClient.cs
```diff
@@ -141,6 +141,15 @@ public sealed class SessionClient : ClientEntity, ISessionClient
         /// </summary>
         public string EntityPath { get; }
 
+        /// <summary>
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -140,6 +140,15 @@ public SubscriptionClient(string connectionString, string topicPath, string subs
         public ReceiveMode ReceiveMode { get; }
 
         /// <summary>
+        /// Duration after which individual operations will timeout.
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/TopicClient.cs
```diff
@@ -91,6 +91,15 @@ public TopicClient(string connectionString, string entityPath, RetryPolicy retry
         public string TopicName { get; }
 
         /// <summary>
+        /// Duration after which individual operations will timeout.
```
SeanFeldman: Would it be better to use `<inheritDoc/>`?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
nemakam: It looks like this tag isn't natively supported by C#. Do I need additional libraries to get this working? I tried using the tag and VS intellisense doesn't show the comment. Am I missing something?

SeanFeldman:
It does 😞
Please ignore my comment.

Issue #240, opened 2017-08-01
MessageHandlerOptions.ExceptionReceivedHandler can be overridden via property setter 


SeanFeldman:
Currently, the following is possible:

```c#
new MessageHandlerOptions(exc =>
{
	Console.WriteLine(exc.ExceptionReceivedContext.Action);
	Console.WriteLine(exc.Exception.Message);
	return Task.FromResult(0);
})
{
	AutoComplete = false, MaxConcurrentCalls = 1,
	ExceptionReceivedHandler = args => Task.FromResult(0)
});
```

This code second assignment of `ExceptionReceivedHandler` will cause logic initially registered with the failure handler to be lost.

`MessageHandlerOptions` should either not allow `ExceptionReceivedHandler` property to be assigned as [handler is mandatory](https://github.com/Azure/azure-service-bus-dotnet/blob/b742d558e37095f66ef9cf1ce4e5c7c3a7c6aefb/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs#L28) (part of constructor defines that) or throw an exception when re-assignment is attempted.

PR #241, opened 2017-08-01
ExceptionReceivedHandler can be set via constructor only to avoid accidental override via getter property

SeanFeldman:
Fixes issue #240

`MessageHandlerOptions` mandates `ExceptionReceivedHandler` argument for the constructor.
At the same time, `ExceptionReceivedHandler` property can be set, leading to an override of the original callback. This PR removes the option to override `ExceptionReceivedHandler` via setter property and forces a valid error callback to be specified during `MessageHandlerOptions` construction time.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
@@ -36,16 +36,13 @@ namespace Microsoft.Azure.ServiceBus
 
         /// <summary>Occurs when an exception is received. Enables you to be notified of any errors encountered by the message pump.
         /// When errors are received calls will automatically be retried, so this is informational. </summary>
-        public Func<ExceptionReceivedEventArgs, Task> ExceptionReceivedHandler { get; set; }
+        public Func<ExceptionReceivedEventArgs, Task> ExceptionReceivedHandler { get; }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
```diff
@@ -36,16 +36,13 @@ public MessageHandlerOptions(Func<ExceptionReceivedEventArgs, Task> exceptionRec
 
         /// <summary>Occurs when an exception is received. Enables you to be notified of any errors encountered by the message pump.
         /// When errors are received calls will automatically be retried, so this is informational. </summary>
-        public Func<ExceptionReceivedEventArgs, Task> ExceptionReceivedHandler { get; set; }
```
vinaysurya: >set; } [](start = 86, length = 6)

Can you please do the same thing for SessionHandlerOptions plz

vinaysurya:
:clock1:

src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
SeanFeldman: Ah, of course! Sorry, completely forgot about it. PR reviews FTW 😃

vinaysurya:
Thanks for the PR @SeanFeldman!

SeanFeldman:
Done @vinaysurya. Also marked the issue itself (#240) as a bug.

PR #244, opened 2017-08-02
Handling session loss for non Session-Lock-Lost exceptions + Increasing AMQP nuget reference to 2.1.0

nemakam:
## Description

Throwing SessionLockLost exception when session is lost due to errors other than SessionLockLostException thrown from the server (eg: communication exceptions).

Also, increased Microsoft.Azure.Amqp nuget reference to 2.1.0

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -728,8 +728,8 @@ namespace Microsoft.Azure.ServiceBus.Core
             ReceivingAmqpLink amqpLink = (ReceivingAmqpLink)sender;
             if (amqpLink != null)
             {
-                AmqpException amqpException = amqpLink.TerminalException as AmqpException;
-                this.LinkError = amqpException != null ? amqpException.Error : null;
+                AmqpException amqpException = amqpLink.GetInnerException() as AmqpException;
+                this.LinkError = amqpException?.Error;
                 MessagingEventSource.Log.SessionReceiverLinkClosed(this.ClientId, this.SessionIdInternal, this.LinkError);
             }
         }
@@ -791,9 +791,9 @@ namespace Microsoft.Azure.ServiceBus.Core
                     a => receiveLink.EndReceiveMessages(a, out amqpMessages),
                     this).ConfigureAwait(false);
 
-                    if (receiveLink.TerminalException != null)
+                    if (receiveLink.GetInnerException() != null)
                     {
-                        throw receiveLink.TerminalException;
+                        throw receiveLink.GetInnerException();
                     }
 
                     if (hasMessages && amqpMessages != null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -791,9 +791,9 @@ void OnSessionReceiverLinkClosed(object sender, EventArgs e)
                     a => receiveLink.EndReceiveMessages(a, out amqpMessages),
                     this).ConfigureAwait(false);
 
-                    if (receiveLink.TerminalException != null)
+                    if (receiveLink.GetInnerException() != null)
                     {
-                        throw receiveLink.TerminalException;
+                        throw receiveLink.GetInnerException();
```
vinaysurya: >GetInnerException [](start = 42, length = 17)

why twice? can we just keep the exception check and throw that exception

nemakam: Changed the logic. Please take a look

src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
```diff
@@ -1210,12 +1211,11 @@ void GetRulesException(string clientId, string exception)
...
 
         [NonEvent]
-        public void CreatingNewLink(string clientId, bool isSessionReceiver, string sessionId, bool isRequestResponseLink, Error linkError)
+        public void CreatingNewLink(string clientId, bool isSessionReceiver, string sessionId, bool isRequestResponseLink, Exception linkException)
         {
             if (this.IsEnabled())
             {
-                string linkErrorString = linkError != null ? linkError.Condition.Value : string.Empty;
-                this.CreatingNewLink(clientId, isSessionReceiver, sessionId ?? string.Empty, isRequestResponseLink, linkErrorString);
+                this.CreatingNewLink(clientId, isSessionReceiver, sessionId ?? string.Empty, isRequestResponseLink, linkException.ToString());
```
vinaysurya: >linkException. [](start = 116, length = 14)

null check?

src/Microsoft.Azure.ServiceBus/MessagingEventSource.cs
```diff
@@ -1226,19 +1226,18 @@ void CreatingNewLink(string clientId, bool isSessionReceiver, string sessionId,
...
 
         [NonEvent]
-        public void SessionReceiverLinkClosed(string clientId, string sessionId, Error linkError)
+        public void SessionReceiverLinkClosed(string clientId, string sessionId, Exception linkException)
         {
             if (this.IsEnabled())
             {
-                string linkErrorString = linkError != null ? linkError.Condition.Value : string.Empty;
-                this.SessionReceiverLinkClosed(clientId, sessionId ?? string.Empty, linkErrorString);
+                this.SessionReceiverLinkClosed(clientId, sessionId ?? string.Empty, linkException.ToString());
```
vinaysurya: >linkException [](start = 84, length = 13)

null check?

PR #245, opened 2017-08-03
MessageReceiver.RenewLockAsync() is not updating the LockedUntilUtc on the Message

vinaysurya:


niemyjski:
To me it only makes sense to take an id  to renew the lock and it's lighter weight. The current implementation needs work and is confusing

PR #275, opened 2017-08-14
MSBuild enforced highest C# language version

SeanFeldman:
## Description

Use MSBuild directive to control the highest version of C# language project permits and do not rely on 3rd party tools such as R#.

<details>
<summary>This following list is used to make sure that common guidelines for a pull request are followed.</summary>

</details>


vinaysurya:
@SeanFeldman, thanks for the PR! Is this primarily to prevent someone who is trying new c# features, not bring them into a repo and throw them a compilation error when they tried to build and submit. Wouldn't an older running compiler anyway catch this and disallow the use of the new feature? Or there are more use cases where this will be able to catch

SeanFeldman:
> The only effect of the langversion switch is to put the compiler in a mode in which use of features from a version of the language higher than the given version cause the compiler to emit an error.

AppVeyor would not necessarily catch C#7.1 if it's not restricted on the build server.
It would also save time and effort for those who are locally running on a high version of C#.

Issue #277, opened 2017-08-16
UWP Scenarios not working with version 1.0.0 library


vinaysurya:
When attempting Universal Windows Platform Send/Receive application using Microsoft.Azure.ServiceBus library which in turn references Microsoft.Azure.Amqp library version 2.1.0, we get the following error during the first send/receive

System.InvalidOperationException: A method was called at an unexpected time. (Exception from HRESULT: 0x8000000E)

The issue has already been fixed in the Azure.Amqp library and a corresponding nuget package is released. Please refer to the following issue for more details on the issue and fix:
https://github.com/Azure/azure-amqp/issues/75

This issue will track the updating of this library version to the newer version of Azure.Amqp library.

eliasj:
Hi, Thank you for the fix. Just what we needed. 

Now I wonder if you know when this fix will be released on NuGet?

nemakam:
@eliasj Unfortunately, we don't have a fixed release date for 1.1.0 yet. We have some more work to do before the release. My best guess for next nuget would be in another 2-3 weeks.

antlhuede:
@nemakam is there a release date for this update coming anytime soon?
Last best guess was for a date several weeks ago. My workaround using this package in UWP is not at all reliable and I don't know if I should spend the time finishing a proper implementation for production or waiting for this package to be released.

Could it be coming with the October 17th Fall Creators Update?

nemakam:
@antlhuede 
Release 2.0 is now out. Please give it a try.

runewake2:
@nemakam The 2.0 release introduces a dependency on .NET Standard 2.0. This requires a minimum version of the Fall Creators Update which makes using this fix very difficult.

nemakam:
@runewake2 
Can you take a look at [2.0 release notes](https://github.com/Azure/azure-service-bus-dotnet/releases/tag/2.0.0) and if you are not particular about any new feature that we released, you could also consider using our 1.0 client with the latest amqp nuget reference in your solution.

PR #278, opened 2017-08-17
Fix to throw ObjectDisposedExceptions when Client is already closed and Update AMQP library version

vinaysurya:
This PR addresses the following

1. Today the client calls fail somewhere internally with messages such as 'X' object already disposed etc. This is changed to consistently throw ObjectDisposed Exceptions when the client is already disposed.
2. Update AMQP Version. Related to issue below:
    -  https://github.com/Azure/azure-service-bus-dotnet/issues/277
3. Some nit updates to order the methods (public>internal>protected>private) correctly (in files touched). We will need to enable the Style rules on this project so we can fix all other places too. Will address this in a different PR.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -115,7 +115,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             int prefetchCount = Constants.DefaultClientPrefetchCount,
             string sessionId = null,
             bool isSessionReceiver = false)
-            : base(ClientEntity.GenerateClientId(nameof(MessageReceiver), entityPath), retryPolicy ?? RetryPolicy.Default)
+            : base(nameof(MessageReceiver), entityPath, retryPolicy ?? RetryPolicy.Default)
         {
             MessagingEventSource.Log.MessageReceiverCreateStart(serviceBusConnection?.Endpoint.Authority, entityPath, receiveMode.ToString());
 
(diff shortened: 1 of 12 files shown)


SeanFeldman:
> We will need to enable the Style rules on this project

@vinaysurya sounds like a new issue 🙂

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1220,36 +1313,35 @@ IList<ArraySegment<byte>> ConvertLockTokensToDeliveryTags(IEnumerable<Guid> lock
...
+                this.LinkException = exception;
+                MessagingEventSource.Log.SessionReceiverLinkClosed(this.ClientId, this.SessionIdInternal, this.LinkException);
+            }
         }
 
-        void CloseRequestResponseSession(RequestResponseAmqpLink requestResponseAmqpLink)
+        IList<ArraySegment<byte>> ConvertLockTokensToDeliveryTags(IEnumerable<Guid> lockTokens)
         {
-            requestResponseAmqpLink.Session.SafeClose();
+            return lockTokens.Select(lockToken => new ArraySegment<byte>(lockToken.ToByteArray())).ToList();
```
SeanFeldman: Returned value is a `List<ArraySegment<byte>>`, should the method signature be `List<...>` as well?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1220,36 +1313,35 @@ IList<ArraySegment<byte>> ConvertLockTokensToDeliveryTags(IEnumerable<Guid> lock
...
         }
 
-        static int ValidateLockTokens(IEnumerable<string> lockTokens)
+        void CloseSession(ReceivingAmqpLink link)
```
SeanFeldman: This method doesn't access any state and is provided with `link`. Why not to make it `static`?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1220,36 +1313,35 @@ IList<ArraySegment<byte>> ConvertLockTokensToDeliveryTags(IEnumerable<Guid> lock
...
-            {
-                throw Fx.Exception.ArgumentNull(nameof(lockTokens));
-            }
-
-            return count;
+            link.Session.SafeClose();
         }
 
-        static int ValidateSequenceNumbers(IEnumerable<long> sequenceNumbers)
+        void CloseRequestResponseSession(RequestResponseAmqpLink requestResponseAmqpLink)
```
SeanFeldman: static?

nemakam:
:shipit:

Issue #269, opened 2017-08-10
How does one configure this library to use a proxy for HTTP traffic?


cjpomer:
In the full framework this was previously handled in web.config or app.config with a default proxy.  Additionally, there is code that can change the default proxy for the process running.  In netcore that functionality is deprecated or unfinished.

But surely there must be a way to consume this API from behind a proxy?

Sorry if this is the wrong forum (I'm a bit new to github).  And I see that other issues are tagged as questions, which this certainly is, but I can't see how/where to do that.

binzywu:
@cjpomer We have not added the websocket support yet which we will add in near future. But we will. However, to give definitive answer about whether proxy is supported across all platforms, we don't know yet.

cjpomer:
Thanks @binzywu.

I want to e.g. receive messages from a queue or topic on a machine running behind a web proxy.

Can you help me understand how websocket will help?

binzywu:
Our underneath [AMQP library](https://github.com/Azure/azure-amqp) supports WebSocket as transportation which uses the WebSocket implementation in .Net which support proxy. WebSocket uses service bus https endpoint.

nemakam:
@cjpomer Websocket support is merged into our branch. This would be shipped as 1.1.0. Make sure you use TransportType=AmqpWebSockets for your connections which run through a proxy.

PR #288, opened 2017-08-21
Added websocket support

NatMarchand:
Added WebSocket support

Should solve the issue #269 

The tests should be runned with a connection string such as : `Endpoint=wss://contoso.servicebus.windows.net;SharedAccessKeyName=keyname;SharedAccessKey=key;`

Unfortunately, the netstandard version lacks the nuget package for websocket support. Shall I had it as a dependency ?

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
@@ -99,6 +99,23 @@ namespace Microsoft.Azure.ServiceBus.Amqp
             return tpSettings;
         }
 
+        public static TransportSettings CreateWebSocketTransportSettings(
+            string networkHost,
+            string hostName,
+            int port)
+        {
+            UriBuilder uriBuilder = new UriBuilder(WebSocketConstants.WebSocketSecureScheme, networkHost, port < 0 ? WebSocketConstants.WebSocketSecurePort : port, WebSocketConstants.WebSocketDefaultPath);
+            WebSocketTransportSettings tcpSettings = new WebSocketTransportSettings
+            {
+                Uri = uriBuilder.Uri,
+                ReceiveBufferSize = AmqpConstants.TransportBufferSize,
+                SendBufferSize = AmqpConstants.TransportBufferSize
+            };
+
+            TransportSettings tpSettings = tcpSettings;
+            return tpSettings;
+        }
+
         public static AmqpConnectionSettings CreateAmqpConnectionSettings(uint maxFrameSize, string containerId, string hostName)
         {
             var connectionSettings = new AmqpConnectionSettings
--- a/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
@@ -42,6 +42,11 @@ namespace Microsoft.Azure.ServiceBus.Primitives
         /// </summary>
         public string SasKeyName { get; set; }
 
+        /// <summary>
+        /// Get the transport type from the connection string
+        /// </summary>
+        public TransportType TransportType { get; set; }
+
         internal FaultTolerantAmqpObject<AmqpConnection> ConnectionManager { get; set; }
 
         public Task CloseAsync()
@@ -54,6 +59,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
             this.Endpoint = new Uri(builder.Endpoint);
             this.SasKeyName = builder.SasKeyName;
             this.SasKey = builder.SasKey;
+            this.TransportType = builder.TransportType;
             this.ConnectionManager = new FaultTolerantAmqpObject<AmqpConnection>(this.CreateConnectionAsync, CloseConnection);
         }
 
@@ -71,11 +77,12 @@ namespace Microsoft.Azure.ServiceBus.Primitives
             AmqpSettings amqpSettings = AmqpConnectionHelper.CreateAmqpSettings(
                 amqpVersion: AmqpVersion,
                 useSslStreamSecurity: true,
-                hasTokenProvider: true);
+                hasTokenProvider: true,
+                useWebSockets: TransportType == TransportType.AmqpWebSockets);
 
             TransportSettings transportSettings = CreateTransportSettings();
-            TransportInitiator initiator = transportSettings.CreateInitiator();
-            TransportBase transport = await ConnectTaskAsync(initiator, timeoutHelper).ConfigureAwait(false);
+            AmqpTransportInitiator initiator = new AmqpTransportInitiator(amqpSettings, transportSettings);
+            TransportBase transport = await initiator.ConnectTaskAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);
 
             string containerId = Guid.NewGuid().ToString();
             AmqpConnectionSettings amqpConnectionSettings = AmqpConnectionHelper.CreateAmqpConnectionSettings(AmqpConstants.DefaultMaxFrameSize, containerId, hostName);
@@ -99,9 +106,8 @@ namespace Microsoft.Azure.ServiceBus.Primitives
             string hostName = this.Endpoint.Host;
             string networkHost = this.Endpoint.Host;
             int port = this.Endpoint.Port;
-            string scheme = this.Endpoint.Scheme;
 
-            if (scheme.Equals(WebSocketConstants.WebSocketSecureScheme, StringComparison.OrdinalIgnoreCase))
+            if (TransportType == TransportType.AmqpWebSockets)
             {
                 return AmqpConnectionHelper.CreateWebSocketTransportSettings(
                     networkHost: networkHost,
@@ -116,26 +122,6 @@ namespace Microsoft.Azure.ServiceBus.Primitives
                     port: port,
                     useSslStreamSecurity: true);
             }
-        }
-
-        private static Task<TransportBase> ConnectTaskAsync(TransportInitiator transportInitiator, TimeoutHelper timeoutHelper)
-        {
-            TaskCompletionSource<TransportBase> tcs = new TaskCompletionSource<TransportBase>();
-            TransportAsyncCallbackArgs callbackArgs = new TransportAsyncCallbackArgs()
-            {
-                CompletedCallback = a =>
-                {
-                    if (a.Exception != null)
-                        tcs.SetException(a.Exception);
-                    else
-                        tcs.SetResult(a.Transport);
-                }
-            };
-
-            if (!transportInitiator.ConnectAsync(timeoutHelper.RemainingTime(), callbackArgs))
-                callbackArgs.CompletedCallback(callbackArgs);
-
-            return tcs.Task;
         }
     }
 }
--- a/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
@@ -66,8 +66,6 @@ namespace Microsoft.Azure.ServiceBus.Primitives
         async Task<AmqpConnection> CreateConnectionAsync(TimeSpan timeout)
         {
             string hostName = this.Endpoint.Host;
-            string networkHost = this.Endpoint.Host;
-            int port = this.Endpoint.Port;
 
             TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
             AmqpSettings amqpSettings = AmqpConnectionHelper.CreateAmqpSettings(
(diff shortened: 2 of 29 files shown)


src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -66,23 +66,16 @@ static void CloseConnection(AmqpConnection connection)
         async Task<AmqpConnection> CreateConnectionAsync(TimeSpan timeout)
         {
             string hostName = this.Endpoint.Host;
-            string networkHost = this.Endpoint.Host;
-            int port = this.Endpoint.Port;
 
             TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
             AmqpSettings amqpSettings = AmqpConnectionHelper.CreateAmqpSettings(
```
nemakam: >CreateAmqpSettings [](start = 61, length = 18)

pass 'true' to `useWebSockets` param here in the case of web sockets.

nemakam:
@NatMarchand , thanks for this contribution. 
Regarding
>The tests should be runned with a connection string such as : Endpoint=wss://contoso.servicebus.windows.net;SharedAccessKeyName=keyname;SharedAccessKey=key;

Our entire test infra would not be easy to change with this type of endpoint. Could you please add couple of tests where you take the endpoint and update the scheme and use it for your tests.

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -100,5 +93,49 @@ static void CloseConnection(AmqpConnection connection)
...
+            {
+                return AmqpConnectionHelper.CreateTcpTransportSettings(
+                    networkHost: networkHost,
+                    hostName: hostName,
+                    port: port,
+                    useSslStreamSecurity: true);
+            }
+        }
+
+        private static Task<TransportBase> ConnectTaskAsync(TransportInitiator transportInitiator, TimeoutHelper timeoutHelper)
```
nemakam: >ConnectTaskAsync [](start = 43, length = 16)

This is pretty much what `AmqpTransportInitiator` does. I do not see a reason why we need to re-write this in our code. Its a lot of code duplication. Could you revert this method and get back to 
`AmqpTransportInitiator initiator = new AmqpTransportInitiator(amqpSettings, tpSettings);`
 `TransportBase transport = await initiator.ConnectTaskAsync(timeoutHelper.RemainingTime()).ConfigureAwait(false);`

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
nemakam: >amqpSettings [](start = 25, length = 12)

While you are at it, could you please pass `true` for `useWebSockets` in this method in the case of web-sockets

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -100,5 +93,49 @@ static void CloseConnection(AmqpConnection connection)
...
         }
+
+        private TransportSettings CreateTransportSettings()
+        {
+            string hostName = this.Endpoint.Host;
+            string networkHost = this.Endpoint.Host;
+            int port = this.Endpoint.Port;
+            string scheme = this.Endpoint.Scheme;
+
+            if (scheme.Equals(WebSocketConstants.WebSocketSecureScheme, StringComparison.OrdinalIgnoreCase))
```
nemakam: >scheme.Equals(WebSocketConstants.WebSocketSecureScheme, StringComparison.OrdinalIgnoreCase) [](start = 16, length = 91)

Unfortunately, this is not how we decide whether to use web socket / tcp. 
Could you add `TransportType` as an enum in `ServiceBusConnectionStringBuilder` which has `Amqp` and `AmqpWebSockets` as the possible options. This has to be parsed from the connection string as a parameter and not by looking at the schema. Which also means, `CSB.ToString()` should appropriately output this parameter when TransportType is not set to default (AMQP).

There are couple of reasons for this call.
1. User might call it https instead of wss
2. (For a most common scenario) The SB namespace connection string from the portal contains sb:// as schema. So to use web sockets, he needs to manually append `TransportType=AmqpWebSockets` or change the schema. So both options require equal amount of work.
3. Changing an enum in `ConnectionStringBuilder` is easier and intuitive than changing the schema.
4. Maintaining compatibility with previous client.

nemakam:
@NatMarchand , you could take a look at EventHub's implementation if required: https://github.com/Azure/azure-event-hubs-dotnet/pull/177

NatMarchand:
I've taken in account all your comments :)
I've not run yet the tests but I'll do it ASAP

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -127,6 +154,15 @@ public string SasKey
...
         }
 
+        /// <summary>
+        /// Get the transport type from the connection string
+        /// </summary>
+        public TransportType TransportType
+        {
+            get => this.transportType;
+            set => this.transportType = value;
+        }
```
nemakam: There's no harm with this code, but this could be just be `{ get; set; }`, and we can get rid of declaring the variable `transportType`

src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
```diff
@@ -99,6 +99,23 @@ internal class AmqpConnectionHelper
             return tpSettings;
         }
 
+        public static TransportSettings CreateWebSocketTransportSettings(
+            string networkHost,
+            string hostName,
+            int port)
+        {
+            UriBuilder uriBuilder = new UriBuilder(WebSocketConstants.WebSocketSecureScheme, networkHost, port < 0 ? WebSocketConstants.WebSocketSecurePort : port, WebSocketConstants.WebSocketDefaultPath);
```
SeanFeldman: >  port < 0 ? WebSocketConstants.WebSocketSecurePort : port

Should this check for a range rather than just the lower boundary?

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -42,6 +42,11 @@ protected ServiceBusConnection(TimeSpan operationTimeout, RetryPolicy retryPolic
         /// </summary>
         public string SasKeyName { get; set; }
 
+        /// <summary>
+        /// Get the transport type from the connection string
```
SeanFeldman: I know this is optional, but having more information on XML doc is such an experience improvement. When intellisense tells you what those are, it save extra operation. Perhaps a \<remarks\> to say AMQP and WebSockets.

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
```diff
@@ -100,5 +100,28 @@ static void CloseConnection(AmqpConnection connection)
...
+            int port = this.Endpoint.Port;
+
+            if (TransportType == TransportType.AmqpWebSockets)
+            {
+                return AmqpConnectionHelper.CreateWebSocketTransportSettings(
+                    networkHost: networkHost,
+                    hostName: hostName,
+                    port: port);
+            }
+            else
```
SeanFeldman: No need in `else`. Code exits if `if` is satisfied.

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -20,8 +20,10 @@ public class ServiceBusConnectionStringBuilder
         static readonly string SharedAccessKeyNameConfigName = "SharedAccessKeyName";
         static readonly string SharedAccessKeyConfigName = "SharedAccessKey";
         static readonly string EntityPathConfigName = "EntityPath";
+        static readonly string TransportTypeConfigName = "TransportType";
```
SeanFeldman: Should these be `const`s instead? (I know this specific is added to what was there already).

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -77,6 +79,31 @@ public ServiceBusConnectionStringBuilder(string endpoint, string entityPath, str
...
+        /// );
+        /// </code>
+        /// </example>
+        /// <param name="endpoint">Fully qualified endpoint.</param>
+        /// <param name="entityPath">Path to the entity.</param>
+        /// <param name="sharedAccessKeyName">Shared access key name.</param>
+        /// <param name="sharedAccessKey">Shared access key.</param>
+        /// <param name="transportType">Transport type</param>
+        public ServiceBusConnectionStringBuilder(string endpoint, string entityPath, string sharedAccessKeyName, string sharedAccessKey, TransportType transportType)
+            : this(endpoint,entityPath,sharedAccessKeyName,sharedAccessKey)
```
nemakam: >, [](start = 27, length = 1)

nit: space

src/Microsoft.Azure.ServiceBus/TransportType.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// AMQP Transport Type
+    /// </summary>
+    public enum TransportType
+    {
+        /// <summary>
+        /// Uses AMQP over TCP
+        /// </summary>
+        Amqp = 0,
```
SeanFeldman: Should it have a remark stating that's the default transport type?

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -127,6 +154,15 @@ public string SasKey
             set => this.sasKey = value.Trim();
         }
 
+        /// <summary>
+        /// Get the transport type from the connection string
+        /// </summary>
+        public TransportType TransportType
```
nemakam: >public TransportType TransportType [](start = 8, length = 34)

Since this is changing, you will have to update the apiappovers.txt as well. Run the apiApprover test and it will generate the new file for you which you need to update.

test/Microsoft.Azure.ServiceBus.UnitTests/ExpectedMessagingExceptionTests.cs
```diff
@@ -12,13 +13,20 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
 
     public class ExpectedMessagingExceptionTests
     {
-        [Fact]
-        async Task MessageLockLostExceptionTest()
+        public static IEnumerable<object> TestConnectionStrings => new object[]
+        {
```
nemakam: I realize you have changed this across all the tests. Sorry for not being more explicit with my testing comment. This changes only the transport layer. Behind the scenes, both transport types will reach the same protocolHeader, and will be parsed exactly same. In short, we do NOT need to test entire functionality again with both TCP and websockets. Only one E2E test which sends a message and receives it should be fine. 
If these were unit tests, i would have left them as-is. But given that these are E2E tests which take quite some time to run, I'll have to ask you to revert these test changes.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
NatMarchand: Well, I have no opinion, I tried to do as it was done in CreateTcpTransportSettings for Tcp

SeanFeldman: @NatMarchand Fully understand. Reply to this question should address both. @nemakam ?

src/Microsoft.Azure.ServiceBus/Primitives/ServiceBusConnection.cs
SeanFeldman: @nemakam ☝️

nemakam:
> I've not run yet the tests but I'll do it ASAP
@NatMarchand Did you get a chance to run all the tests?

NatMarchand:
Just did, works like a charm :)

nemakam:
@NatMarchand , The test that you added seem to be failing - https://ci.appveyor.com/project/vinaysurya/azure-service-bus-dotnet/build/job/q30ld3m1fkn72qk4/tests. See if you can take a look and find out what's wrong.

src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
nemakam: Sorry, just noticed this. If the user has given an out of range port, then he is either trying to fool around the system or has a wrong connection string in the first place. In either of the cases, he will get an exception from the lower layer and it won't harm us. But the plus point is that for the second scenario he might realise something is wrong and double check his connection string. We shouldn't be defaulting to our 443 when he is giving a out of range port on purpose. This check for <0 is useful since if no port details are given, UriBuilder returns a default of -1, and that is when we may have to apply our default port.
So in short, the current code looks right to me

---
In reply to: [134849248](https://github.com/Azure/azure-service-bus-dotnet/pull/288#discussion_r134849248) [](ancestors = 134849248)

NatMarchand:
Ah my bad, it's missing WebSocket package reference for netcoreapp runtime in the test project.

`<PackageReference Include="System.Net.WebSockets.Client" Version="4.0.0" />`
Do I need to make an other pull request?

SeanFeldman:
The test project should pull in all the main project dependencies by
default iirc. Might be a nuget issue.

NatMarchand:
From what I have seen in the eventhub dotnet sdk, only the test project has the reference, not the src project.
IMHO the package Microsoft.Azure.Amqp should reference System.Net.WebSockets.Client (the csproj does, not the nuget package)

nemakam:
@NatMarchand , I agree. Meanwhile, could you please update our test csproj to include this dependency so that your test starts running?

nemakam:
Raised a new issue in amqp library for this. https://github.com/Azure/azure-amqp/issues/76

PR #304, opened 2017-09-08
Add a helpful message about MaxAutoRenewDuration and permissions

paulgmiller:
If you're using a listen sas it's extremely hard to debug why auto renew isn't working

## Description

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
@@ -62,7 +62,8 @@ namespace Microsoft.Azure.ServiceBus
         public bool AutoComplete { get; set; }
 
         /// <summary>Gets or sets the maximum duration within which the lock will be renewed automatically. This
-        /// value should be greater than the longest message lock duration; for example, the LockDuration Property. </summary>
+        /// value should be greater than the longest message lock duration; for example, the LockDuration Property. 
+        /// Also client must have manage rather than just listen permissions (no error will be thrown)</summary>
         /// <value>The maximum duration during which locks are automatically renewed.</value>
         public TimeSpan MaxAutoRenewDuration
         {


src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
```diff
@@ -62,7 +62,8 @@ public int MaxConcurrentCalls
         public bool AutoComplete { get; set; }
 
         /// <summary>Gets or sets the maximum duration within which the lock will be renewed automatically. This
-        /// value should be greater than the longest message lock duration; for example, the LockDuration Property. </summary>
+        /// value should be greater than the longest message lock duration; for example, the LockDuration Property. 
+        /// Also client must have manage rather than just listen permissions (no error will be thrown)</summary>
```
SeanFeldman: Sounds like a \<remark\>. With a remark you would not need the "also".

Additional info could be "renewal is not a guaranteed operation and can fail".

nemakam:
Marking this as hold for now.
We need to evaluate if we need to change the behavior or add documentation for this.

nemakam:
And regarding
> If you're using a listen sas it's extremely hard to debug why auto renew isn't working

I just tried repro'ing the scenario. I see that we do throw an exception in this case which clearly mentions that `Send` claim is required for this operation. I would suggest you to listen and log all the exceptions received using the exception received handler.

paulgmiller:
Interesting we just added an exceptionhandler at the same time we had problems here but I think it got lost in other noisy instrumentation for our integration beds (obviously our fault). 
Looking back at the instrumentation I see MessageLockLostException with 
"The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue."

But I don't see any thing about renewal requiring a send claim.

nemakam:
@paulgmiller `Manage` claim created on portal contains both send and receive implicitly. Retry your experiment with only listen claim and it should show you the error.

src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
SeanFeldman: @paulgmiller ping

nemakam:
Closing this in favor of #369 
They both are due to the same issue.

Issue #190, opened 2017-06-16
Running tests on PRs w/o jeopardizing security


SeanFeldman:
Currently PRs from external contributors are not running tests on build server (AppVeyor).

This creates a few problem of
- Contributors not being aware if they break anything by their changes
- ASB team members required to run those tests locally/manually

It would be ideal to find out a way to to execute tests on PRs as well.

/cc @nemakam

SeanFeldman:
One of the discussed options is when a build is executed, it create a resource group `sb-dotnet-av-$buildVersion-rg` which is removed when the build is done. This means PRs would be creating their own RGs and removing those when the build is done. 

Problem with this approach is that it makes secrets `azure-service-bus-dotnet/ClientSecret` and `azure-service-bus-dotnet/CodeCovSecret` vulnerable 😞

SimonCropp:
only build PRs that have had a review form someone in the team

 * configure appveyor to not build any public PR branches. 
 * Build small bot
 * bot detects a PR "approval" either by interrogating the approval state of a PR https://developer.github.com/v3/pulls/reviews/#list-reviews-on-a-pull-request or by using a custom label.
 * note the bot run periodical or respond to webhooks like https://developer.github.com/v3/activity/events/types/#pullrequestreviewevent https://developer.github.com/v3/activity/events/types/#pullrequestevent
 * ping appveyor to start a build for that specific branch https://www.appveyor.com/docs/api/projects-builds/#start-build-of-specific-branch-commit

nemakam:
I tried a different approach. Let me know if you find any issues with it.

* Blocked appveyor from being configured from `appveyor.xml`. Instead, moved everything to UI based approach where I have given the `build.ps1` inline as the build script.
* The script reads the environment variables into local variables, and then before invoking MsBuild, go and update environment variables with garbage value.
* Use the locally read variables to take care of test bed setup

nemakam:
Looks like this works for now.

PR #309, opened 2017-09-10
Renew locktoken without entire message

### LINKED ISSUE #295: Provide RenewLockAsync functionality which accepts only LockToken
IMessageReceiver method:
 Task RenewLockAsync(Message message);
Takes "Message" as the argument as opposed to all other similar methods which takes" string lockToken" as follows:

Task CompleteAsync(IEnumerable<string> lockTokens);
Task DeferAsync(string lockToken);
 Task AbandonAsync(string lockToken);
 Task DeadLetterAsync(string lockToken);

Why RenewLockAsync is taking Message object and not lockToken string as the argument?

How do you expect the user to use the RenewLockAsync method when we don't expose the Message object to them?

On search I found this issue and understood the reason, however, what is the way to create Message Object given we have lock token?
https://github.com/Azure/azure-service-bus-dotnet/pull/245

SeanFeldman:
Fixes #295

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
@@ -125,6 +125,19 @@ namespace Microsoft.Azure.ServiceBus.Core
         Task RenewLockAsync(Message message);
 
         /// <summary>
+        /// Renews the lock on the message. The lock will be renewed based on the setting specified on the queue.
+        /// <returns>New lock token expiry date and time in UTC format.</returns>
+        /// </summary>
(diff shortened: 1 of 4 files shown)


nemakam:
(y)

nemakam:
@SeanFeldman, wonder why this didnt require you to update the approvedApi list

SeanFeldman:
Related to #190 :disappointed:
Therefore `APIApprovals` [wasn't executed](https://ci.appveyor.com/project/vinaysurya/azure-service-bus-dotnet/build/1.0.288/job/fuvqb14eith02acu) by the build server.
Pushed updated @nemakam

Issue #294, opened 2017-09-03
MessageReceiver.ReceiveAsync throws ServiceBusTimeoutException if passed with Timespan.Zero


shahbarkha:
In your sample application :\azure-service-bus\samples\DotNet\Microsoft.Azure.ServiceBus\SendReceiveUsingMessageSenderReceiver

Please try to pass Timespan.Zero in the ReceiveAsync method. 

Actual Behavior - Exception gets thrown. 

An unhandled exception of type 'Microsoft.Azure.ServiceBus.Primitives.ServiceBusTimeoutException' occurred in System.Private.CoreLib.ni.dll
Timed out trying to create FaultTolerantAmqpObject`1

Expected Behavior - Should it not retrun null? What does it mean to pass TimeSpan.Zero in the ReceiveAsync method?

Message message = await messageReceiver.ReceiveAsync(TimeSpan.Zero);

SeanFeldman:
`null` would be returned if the invoked `this.ReceiveAsync(1, operationTimeout)` would not fail. But it fails with the underlying exception is  `Timed out trying to create FaultTolerantAmqpObject`1`. That's because `ReceiveAsync(int maxMessageCount, TimeSpan operationTimeout)` is subsequently invoked with `TimeSpan.Zero` for the `operationTimeout`. That one is passed further into `this.OnReceiveAsync()` call. When [`ReceiveLink`](https://github.com/SeanFeldman/azure-service-bus-dotnet/blob/b4285e5b436f0574810c49cacdd9973e214d155c/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L789) is created, it throws as timeout has already occurred (timespan of zero).

I'd say `ReceiveAsync()` should not accept `TimeSpan.Zero`.

niemyjski:
@SeanFeldman, I agree. Any timespan value should be properly handled with argument validation or from the service. TimeSpan.Zero is a valid value saying I do not wish to timeout. If it's not supported server side an ArgumentException should be the proper course.

@barkhashah perhaps if our implementation contains should check for TimeSpan.Ticks <= 0 and not pass a timeout value? @SeanFeldman, do you know off the top of your head if you even pass this through if it's negative? Perhaps it's just ignored if it's 0 or negative timespan, but the counter argument is fail fast.

SeanFeldman:
[...]

PR #313, opened 2017-09-12
Adding additional documentation + typos

### LINKED ISSUE #311: Is "Authentican" a spelling error?
https://github.com/Azure/azure-service-bus-dotnet/search?utf8=%E2%9C%93&q=Authentican&type=

nemakam:
Fixes #311 and sets right expectation for #294

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
@@ -52,6 +52,11 @@ namespace Microsoft.Azure.ServiceBus.Core
         /// </summary>
         /// <param name="operationTimeout">The time span the client waits for receiving a message before it times out.</param>
         /// <returns>The message received. Returns null if no message is found.</returns>
+        /// <remarks>
+        /// The parameter <paramref name="operationTimeout"/> includes the time taken by the receiver to establish a connection
+        /// (either during the first receive or when connection needs to be re-established). If establishing the connection
+        /// times out, this will throw <see cref="ServiceBusTimeoutException"/>.
+        /// </remarks>
         Task<Message> ReceiveAsync(TimeSpan operationTimeout);
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
```diff
@@ -119,8 +128,8 @@ public interface IMessageReceiver : IReceiverClient
         /// <remarks>
         /// When a message is received in <see cref="ServiceBus.ReceiveMode.PeekLock"/> mode, the message is locked on the server for this
         /// receiver instance for a duration as specified during the Queue/Subscription creation (LockDuration).
-        /// If processing of the message requires longer than this duration, the lock needs to be renewed. For each renewal, the lock is renewed by
-        /// the entity's LockDuration.
+        /// If processing of the message requires longer than this duration, the lock needs to be renewed.
+        /// For each renewal, it resets the time the message is locked by the LockDuration set on the Entity.
```
SeanFeldman: I would even add that this is **not** a guaranteed operation and can fail, resulting in message being unlocked.

nemakam: Personally, I'm not in favor of having that in the documentation. If the call `await queueClient.RenewMessageLock(lockToken)` executes successfully without any exception, then the lock must have been renewed. Else, we would throw an exception (say in scenarios where lock has already expired). Throwing the exception is already a clear indication that it did not work.

SeanFeldman: @nemakam receiving an exception at _run-time_ is a clear indication that something didn't work. Reading API doesn't not let you know it _can_ happen. You can be running this command a thousand times and it will be successful, but on a thousand and one it will throw an exception.

nemakam: @SeanFeldman, sure it can fail. I'm not denying the possibility of failing. 
And this operation failing wouldn't result in message being unlocked. Its the other way. Message being unlocked would result in this operation failing. That's true with almost any operation that we have. Don't think we need to specify that with every operation, that it can fail.

nemakam: And moreover, the entire description of this method should make that clear. That if you need to spend more time with your message, then you need to renew the lock, else it will expire. The concept of expiration is not something that we may have to explain again in the same summary.

src/Microsoft.Azure.ServiceBus/Primitives/DispositionStatus.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
     enum DispositionStatus
     {
         Completed = 1,
-        Deferred = 2,
+        Defered = 2,
```
SimonCropp: perhaps add a comment here for the next person

PR #310, opened 2017-09-11
Cleanup

SimonCropp:
also fixes https://github.com/Azure/azure-service-bus-dotnet/issues/301

please cherry pick or merge as you want

PR #320, opened 2017-09-13
Cleanup v3

SimonCropp:


SimonCropp:
can i get this reviewed/merged/cherrypicked/declined given i have another one coming that will conflict if i do it after

nemakam:
@SimonCropp Currently trying to get #190 working so that appveyor build can run on this PR before we check it in. Any issues in the code should be directly found before checkin. Don't want to repeat issues like those in #310, some of which cannot be caught easily on dev machine.
If you have suggestions on how to get #190 done without losing security, let us know!

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -916,13 +931,13 @@ public override void UnregisterPlugin(string serviceBusPluginName)
             }
         }
 
-        protected virtual async Task<IList<Message>> OnReceiveDeferredMessageAsync(IEnumerable<long> sequenceNumbers)
+        protected virtual async Task<IList<Message>> OnReceiveDeferredMessageAsync(long[] sequenceNumbers)
```
SimonCropp: I know i changed a public API here. but having an IEnumerable on a public api that is also in the fast path is problematic.

Given you dont know the producer of the IEnumerable, you dont know if it supports multiple iterations, or if it s a consume and discard approach. as such u need to play it safe and covert it to a local list or array, prior to doing multiple enumerations. You might want to consider making these types of APIs and IList, Icollection or an array

nemakam: Valid point :thumbsup: 

---
In reply to: [138780098](https://github.com/Azure/azure-service-bus-dotnet/pull/320#discussion_r138780098) [](ancestors = 138780098)

src/Microsoft.Azure.ServiceBus/Primitives/TokenProviderAdapter.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
+
```
nemakam: nit: move it inside namespace

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
SimonCropp: @nemakam so given there are other `IEnumerable`s in the public API how would u like to handle those?

nemakam: @SimonCropp  Anything which is not part of interface, we could attempt at changing it. I don't want to change anything that's already part of interface. We should solve them by doing `.ToList()` and use them

nemakam:
@SimonCropp Looks like you need to update API approval.txt again.

SimonCropp:
@nemakam yep forgot. done

Issue #67, opened 2017-01-18
Path and Name confusion


SeanFeldman:
This is more of a question/discussion issue.
With the old client, there was somewhat of a confusion around Path and Name.

#### Queues

`this/is/a/queue`
[path](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.queuedescription#Microsoft_ServiceBus_Messaging_QueueDescription_Path): `this/is/a/queue` 

#### Topics
`a/b/c`
[path](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.topicdescription#Microsoft_ServiceBus_Messaging_TopicDescription_Path): `a/b/c`

#### Subscription
`some/topic/subscriptions/sub`
[topic path](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.subscriptiondescription#Microsoft_ServiceBus_Messaging_SubscriptionDescription_TopicPath): `some/topic`
[name](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.subscriptiondescription#Microsoft_ServiceBus_Messaging_SubscriptionDescription_Name): `sub`

While queues and topics share a common `Path` concept, they are not consistent with subscriptions. Would it helpful to have a `FullPath` abstraction that would be available for all 3 entity types?

Queues: `FullPath` == `Path`
Topics: `FullPath` == `Path`
Subscriptions: `FullPath` == `TopicPath`/subscriptions/`Name`

jtaubensee:
@SeanFeldman - What you are saying makes sense. Would it be more clear if we just provided the name and got rid of the path all together? Is there any scenario you can think of where the path is particularly important?

SeanFeldman:
Path makes more sense to me tbh as it can have segments (separated by `/`).
The outlier is subscriptions. Since they have a segment that breaks that pattern therefore there's a name that makes sense. I'd like to keep a name for subscriptions since at the application level you refer to a subscription by its name, not path. Hence the suggest to have a `FullPath`. What does the team say?

abombss:
Not just subscriptions but $deadletter as well. Would love to see the full path concept instead of name. Personally I hate having to deal with queues, topics, and subscriptions differently. They are just senders or receivers in my book with an address.

jtaubensee:
Hey Guys - Just doing some cleanup here. 

[...]

PR #324, opened 2017-09-15
cleanup v4

SimonCropp:


src/Microsoft.Azure.ServiceBus/QueueClient.cs
```diff
@@ -78,7 +78,6 @@ public QueueClient(ServiceBusConnectionStringBuilder connectionStringBuilder, Re
         /// Instantiates a new <see cref="QueueClient"/> to perform operations on a queue.
         /// </summary>
         /// <param name="connectionString">Namespace connection string. Must not contain queue information.</param>
-        /// <param name="entityPath">Name of the queue</param>
```
nemakam: >/// <param name="entityPath">Name of the queue</param> [](start = 8, length = 54)

I think this should be left as is. Revert this. Not everyone might understand the 'path' here

SeanFeldman: @nemakam, isn't that wrong? if the parameter is "entity path", it can't be an "entity name".

`org/department/queue` for example, the **path** is `org/department/queue`, but **name** is `queue`

If it's reverted, then the change should be
```
/// <param name="entityPath">Path of the queue</param>
```

nemakam: `org/department/queue`.. What scenario is this? what is `org` and `department` in this case?
For us, `queue` doesn't really have a path. It just has a name. Even `topic` has a name. But entities like `subscription` have paths.
So I would call `org/department/queue` as the queue *name*.

SimonCropp:
Then shouldnt the parameter name be queueName?

src/Microsoft.Azure.ServiceBus/QueueClient.cs
SeanFeldman: This brings back [this issue](https://github.com/Azure/azure-service-bus-dotnet/issues/67#issuecomment-298722100) and what was describes back in May.

I think it will be confusing to see

https://github.com/Azure/azure-service-bus-dotnet/blob/364b48319987421b3e931cd63b048966391bd705/src/Microsoft.Azure.ServiceBus/QueueClient.cs#L121

and 

https://github.com/Azure/azure-service-bus-https://github.com/Azure/azure-service-bus-dotnet/blob/364b48319987421b3e931cd63b048966391bd705/src/Microsoft.Azure.ServiceBus/QueueClient.cs#L140

at the same time.

nemakam: If you look at the implementation of `QueueClient.Path`, we just do `return this.QueueName`. Now it brings us to, why have both? Coz we have a common parameter `Path` in every client entity (`MessageReceiver` to `QueueClient` to `SubscriptionClient` etc). But now, we have a `QueueClient` as a separate class which is aimed at people who potentially might not have a complete understanding of what `Path` means? Since this is a client specific to queues, we have a `QueueName` as an additional property. This property is especially required when you start thinking of say `SubscriptionClient` which needs to have two such properties `SubscriptionName` and `TopicName` along with the usual `Path` which are all different. Its just about having consistency across the board.

nemakam:
@SimonCropp , now to think of it, don't want to change public APIs unnecessarily when it's not required. I'm sure there are lot of people using named parameters in their code. I don't think this change is worth breaking all their code. Lets stick to `entityPath` with param xml doc mentioning its the queue name

SimonCropp:
@nemakam feedback addressed

PR #325, opened 2017-09-16
Closing amqp-session when creating link fails.

### LINKED ISSUE #237: QuotaExceededException
I started my console app two days ago, and have since then received no more than 4 messages. Yet today, I get this exception:

> Session handler exception "Microsoft.Azure.ServiceBus.QuotaExceededException: Cannot allocate more handles. The maximum number of handles is 4999.

I'm not sure what kind of handles it is referring to, but to imagine that I would have exceeded some form of quota after this little activity seems wrong.

nemakam:
## Description

We create a new amqp-session on which we open links. But if link creation fails, we should close the session as it is not re-used again. These sessions are mapped to a session handle table on backend which keeps track of these open sessions. Once we hit 5000 open sessions, backend throws QuotaExceededException.
Fixes #237

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpLinkCreator.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpLinkCreator.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
                     amqpConnection,
                     exception);
 
-                session.Abort();
+                session.SafeClose(exception);
                 throw AmqpExceptionHelper.GetClientException(exception, null, link?.GetInnerException(), session.IsClosing());
             }
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.ServiceBus/Amqp/AmqpLinkCreator.cs
```diff
@@ -78,6 +78,7 @@ protected AmqpLinkCreator(string entityPath, ServiceBusConnection serviceBusConn
                     amqpConnection,
                     exception);
 
+                session.Abort();
```
xinchen10: Call SafeClose instead. Abort does not notify the remote peer of state change.

nemakam: @xinchen10  So should it be updated for exceptions while creating session? In line 61 above

Issue #335, opened 2017-09-24
Make previously available classes/enums public again


alexeyzimarev:
I spent some time working on the MassTransit migration to use this library.

However, many elements are now made internal or private, whilst we count on them to be public.

Examples:

- `TokenProvider` - internal abstract static class but static methods are public
- `TokenScope` - internal enum

nemakam:
@alexeyzimarev 
I would like to know your use-case for this request. 
We already have a request to allow using SAS tokens (#332). Does this take care of your scenario?

alexeyzimarev:
@nemakam as I wrote, I am working to switch MassTransit from the old ServiceBus library to this library. The API I mentioned was public before. It means that in the new library the API surface has changed so much that we now need to seek workarounds for something that just worked before.

I give you an example about `TokenProvider`. This class has several static methods to create a shared access key token provider with different set of parameters. *None* of those methods are being used in the library itself. What is the reason this API has even been made? Definitely, it would be useful for the API users to have the ability to call these methods. By doing so, you will not force us to support all these ways to create the shared access key token provider by all parameters combinations that you accept, but just let our users to choose, which factory method to use from the `TokenProvider` class itself and let us keep the reference to the `TokenProvider` object.

alexeyzimarev:
I mean, how do we suppose to use these methods:
https://github.com/Azure/azure-service-bus-dotnet/blob/10fb579c967200563c108b4e4282de5ead79730e/src/Microsoft.Azure.ServiceBus/Primitives/TokenProvider.cs#L42
https://github.com/Azure/azure-service-bus-dotnet/blob/10fb579c967200563c108b4e4282de5ead79730e/src/Microsoft.Azure.ServiceBus/Primitives/TokenProvider.cs#L53
https://github.com/Azure/azure-service-bus-dotnet/blob/10fb579c967200563c108b4e4282de5ead79730e/src/Microsoft.Azure.ServiceBus/Primitives/TokenProvider.cs#L81
https://github.com/Azure/azure-service-bus-dotnet/blob/10fb579c967200563c108b4e4282de5ead79730e/src/Microsoft.Azure.ServiceBus/Primitives/TokenProvider.cs#L93

nemakam:
Yes. These need to get public. These will be made public as part of #332. 
Meanwhile If you are blocked on this, it would be best if you could raise a PR for this. I can help you out with the changes that are to be done.

nemakam:
[...]

PR #337, opened 2017-09-28
Making some internal classes and enums public

alexeyzimarev:
Making some internal classes and enums public.
https://github.com/Azure/azure-service-bus-dotnet/issues/335

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Primitives/SecurityToken.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/SecurityToken.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
     /// <summary>
     /// Provides information about a security token such as audience, expiry time, and the string token value.
     /// </summary>
-    internal class SecurityToken
+    public class SecurityToken
     {
         // per Simple Web Token draft specification
         private const string TokenAudience = "Audience";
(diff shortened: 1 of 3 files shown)


nemakam:
@alexeyzimarev , Thanks for the changes, but I don't think this is sufficient enough. You still cant use them even if you make these classes public since there is no way to provide a SAS token while creating a connection to service bus in this client.
This requires several things to be changed:
1. Instead of entities doing `var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider()`, this needs to come from the `ServiceBusConnectionStringBuilder`.
1. In order to do that we will have to expose properties on `ServiceBusConnectionStringBuilder` which are required (mandatory or optional params) to create the token. E.g, `SharedAccessSingature` (SASToken), `TokenTimeToLive`, `TokenScope`.
1. Additionally we could also expose `TokenProvider` in connectionstring builder for people to plug in their own providers.
1. Given that we are already providing `TokenProvider` as a property, why should we still add properties like `SharedAccessSignature`? Since we support it natively and we would like to have `connectionStringBuilder.ToString()` return a connectionString which has `SharedAccessSignature=<>` in the string.
1. We will have to update all the entities to read `connectionStringBuilder.TokenProvider` instead of creating a new TokenProvider in the constructor.

alexeyzimarev:
I think Clemens already made necessary changes in #338. I will wait for it to be merged and check if I can proceed. At least, some of the requested changes are made there.

SeanFeldman:
Should this be closed @nemakam?

vinaysurya:
Fixed with https://github.com/Azure/azure-service-bus-dotnet/pull/385

PR #338, opened 2017-09-29
Added SharedAccessSignature support for Connection String

### LINKED ISSUE #332: Support a Shared Access Signature to create a Queue|Topic|SubscriptionClient
## Actual Behavior
1. All the clients have constructors that accept a ServiceBusConnectionStringBuilder or a connectionString that creates a new ServiceBusConnectionStringBuilder.
2. ServiceBusConnectionStringBuilder seems that not support Shared Access Signatures
3. The result is that all TokenProvider instances are created based on the serviceBusConnection.SasKeyName and the serviceBusConnection.SasKey
4. It forces the client to know that sensitive information.

## Expected Behavior
1. ServiceBusConnectionStringBuilder should support Shared Access Signatures
2. The TokenProvider should be created with the existing factory method: TokenProvider.CreateSharedAccessSignatureTokenProvider(serviceBusConnection.SharedAccessSignature)

## Versions  
- OS platform and version: All
- .NET Version: All
- NuGet package version or commit ID: 1.0.0

clemensv:
Added SharedAccessSignature option for connection string builder, moved control of creating token provider into ServiceBusConnection,  added test. 

This will also resolve #332

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
 
             this.ownsConnection = true;
-            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(this.ServiceBusConnection.SasKeyName, this.ServiceBusConnection.SasKey);
+            var tokenProvider = this.ServiceBusConnection.CreateTokenProvider();
             this.CbsTokenProvider = new TokenProviderAdapter(tokenProvider, this.ServiceBusConnection.OperationTimeout);
         }
 
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -74,9 +74,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
 
             this.ownsConnection = true;
-            var tokenProvider = TokenProvider.CreateSharedAccessSignatureTokenProvider(
-                this.ServiceBusConnection.SasKeyName,
-                this.ServiceBusConnection.SasKey);
+            var tokenProvider = this.ServiceBusConnection.CreateTokenProvider();
             this.CbsTokenProvider = new TokenProviderAdapter(tokenProvider, this.ServiceBusConnection.OperationTimeout);
         }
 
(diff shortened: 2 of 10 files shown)


nemakam:
@clemensv , Thanks for the changes.
IMO there should be more to this than what is currently implemented. I've explained parts of it [here](https://github.com/Azure/azure-service-bus-dotnet/pull/337#issuecomment-332925912). 

We should expose some way of creating the SasToken. Either we expose `TokenProvider` as a property, or expose utility methods to generate SASToken from `KeyName` and `AccessKey`.

alexeyzimarev:
Or both

nemakam:
I cant think of a scenario for which we need to expose `TokenProvider` as a settable property. The main scenario is to be able to generate a `SasToken`. 
So we could probably enable this scenario by allowing this:
`string sasToken = TokenProvider.CreateSharedAccessSignatureTokenProvider(<key>, <value>, <ttl>, <scope>).BuildSasToken()'

nemakam:
A slightly cleaner approach would be to make `SharedAccessSignatureTokenProvider` and `TokenProvider` public with all its methods changed to internal. And then expose `GenerateSasToken(key, value, resource, ttl)` on SasTokenProvider class as a static method. 
We don't really need to expose `TokenProvider.CreateSharedAccessSignatureTokenProvider()`. This will also help us keep the design clean in case we decide later to support other kind of token providers other than SAS.

alexeyzimarev:
It might not be required for us, what I actually need is to have an API that allows me to use both connection string and individual settings. This is why I think that having the connection string builder that can handle this might work. So I will just always use the connection string, either a complete one, or the one I build from individual settings.

But to see how it is used now, you can look at: https://github.com/MassTransit/MassTransit/blob/develop/src/MassTransit.AzureServiceBusTransport/ServiceBusHost.cs#L430

nemakam:
@alexeyzimarev 
> having the connection string builder that can handle this might work. So I will just always use the connection string, either a complete one, or the one I build from individual settings.

If I understand your use case correctly, most of it is already done with this PR. The only scenario which I think will be missing is to provide `sasKeyName` and `sasKeyValue` and being able to provide your own TTL. The way to solve this would be is to generate a SASToken and then pass that to the string builder. 

Why I'm not personally in favor of exposing TTL and TokenScope in string builder is that it doesn't solve any scenario that I can think of. My assumption here is that once you have SasKeyName and SasKeyValue, there is no reason to allow manipulation of TTL and scope. You could let it have the default values.

alexeyzimarev:
Yes, this is what I thought too: if this gets merged, I can continue, at least for some time.
I can change the API to remove the TTL and token scope. In fact I only need to be able to supply either a full connection string, or key name/value to create the client.

clemensv:
I agree that we should expose the TokenProvider (again as in NETFX), but this PR unblock being able to use a token at least from the connection string. Baby steps.

nemakam:
@clemensv , Looks like the approvedAPI needs to be updated again

alexeyzimarev:
Any news on this?

PR #340, opened 2017-10-03
Spelling

SeanFeldman:
Minor fix

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Primitives/DispositionStatus.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/DispositionStatus.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
     enum DispositionStatus
     {
         Completed = 1,
-        Defered = 2,
+        Deferred = 2,
         Suspended = 3,
         Abandoned = 4,
         Renewed = 5,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.ServiceBus/Primitives/DispositionStatus.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
     enum DispositionStatus
```
nemakam: Check out the comments on top of this enum.

SimonCropp:
@nemakam so where is the plan fix this long term?

nemakam:
@SimonCropp , we have that as a backlog item in our service side tasks. Once that is changed, we will update all the clients. However, for backward compatibility, it would be difficult to stop supporting the typo'd version.

SimonCropp:
> However, for backward compatibility, it would be difficult to stop supporting the typo'd version.

true. but you could also support the correct spelling moving forward

SeanFeldman:
Adding more descriptive comment might add clarity.​

SimonCropp:
>  we have that as a backlog item in our service side tasks. Once that is changed, we will update all the clients

can u share the url for that issue

nemakam:
> can u share the url for that issue

We track and plan our work in VSTS (VisualStudio.com) and not github. Unfortunately, it doesn't support anonymous access.

SimonCropp:
@nemakam then you should be clear about that and remove the issues tab from this github repo.

PR #341, opened 2017-10-04
Added overload for deadletterAsync to specify dead letter reason

nemakam:
Added overload for deadletterAsync to specify dead letter reason.
Resolves #339

PR #342, opened 2017-10-06
Inlined variables

SeanFeldman:
Connects to #341 

- Inlining variables

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -1453,15 +1453,18 @@ namespace Microsoft.Azure.ServiceBus.Core
                     rejected.Error.Info.Add(Message.DeadLetterErrorDescriptionHeader, deadLetterErrorDescription);
                 }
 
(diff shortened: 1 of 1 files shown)


nemakam:
Let me trigger a new build after merging with #347

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1455,19 +1453,15 @@ Rejected GetRejectedOutcome(IDictionary<string, object> propertiesToModify, stri
                     rejected.Error.Info.Add(Message.DeadLetterErrorDescriptionHeader, deadLetterErrorDescription);
                 }
 
-                if (propertiesToModify != null)
+                foreach (var pair in propertiesToModify)
```
nemakam: This has to be wrapped inside `if (propertiesToModify != null)`

SeanFeldman: Silly mistake, sorry. Fixing.

SeanFeldman:
@nemakam done.

PR #349, opened 2017-10-09
Updated inline comments for message properties

clemensv:
Updated the XML comments for message properties to align with the newly published docs.


src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -53,8 +53,14 @@ public Message(byte[] body)
         /// <summary>
         /// Gets or sets the MessageId to identify the message.
         /// </summary>
-        /// <remarks>A value set by the user to identify the message. In case message deduplication is enabled on the entity, this value will be used for deduplication.
-        /// Max MessageId size is 128 chars.</remarks>
```
SeanFeldman: Is size limit on MessageId when de-dup is on no longer in place?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -66,10 +72,14 @@ public string MessageId
             }
         }
 
-        /// <summary>Gets or sets a partition key for sending a transactional message to a queue or topic that is not session-aware.</summary>
-        /// <value>The partition key for sending a transactional message.</value>
-        /// <remarks>Transactions are not currently supported with this library. Messages with same partitionKey are sent to the same partition.
-        /// Max PartitionKey size is 128 chars.</remarks>
+        /// <summary>Gets or sets a partition key for sending a message to a partitioned entity.</summary>
+        /// <value>The partition key. Maximum length is 128 characters.</value>
```
SeanFeldman: [Documentation on partitioning](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning) doesn't contain the max length info. Should it?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -169,25 +199,48 @@ public TimeSpan TimeToLive
...
+        /// </remarks>
         public string CorrelationId { get; set; }
 
-        /// <summary>Gets or sets the application specific label.</summary>
-        /// <value>The application specific label.</value>
+        /// <summary>Gets or sets an application specific label.</summary>
+        /// <value>The application specific label</value>
+        /// <remarks>
+        ///   This property enables the application to indicate the purpose of the message to the receiver in a standardized 
+        ///   fashion, similar to an email subject line. The mapped AMQP property is "subject".
```
SeanFeldman: Would be nice to have a link to a document that describes these mappings.

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -227,8 +280,8 @@ public override string ToString()
             return string.Format(CultureInfo.CurrentCulture, $"{{MessageId:{this.MessageId}}}");
         }
 
-        /// <summary>Clones a message, so that it is possible to send a clone of a message as a new message.</summary>
-        /// <returns>The <see cref="Message" /> that contains the cloned message.</returns>
+        /// <summary>Clones a message, so that it is possible to send a clone of an already received message as a new message.</summary>
```
SeanFeldman: System properties are [not cloned](https://github.com/Azure/azure-service-bus-dotnet/blob/468d02fd16a314ba967662fe382e3457c5b1e1f7/src/Microsoft.Azure.ServiceBus/Message.cs#L245). Should that be added as a remark?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -294,16 +347,24 @@ public sealed class SystemPropertiesCollection
             /// <summary>
             /// Gets the lock token for the current message.
             /// </summary>
-            /// <remarks>A lock token will only be specified if the message was received using <see cref="ReceiveMode.PeekLock"/></remarks>
+            /// <remarks>
+            ///   The lock token is a reference to the lock that is being held by the broker in <see cref="ReceiveMode.PeekLock"/> mode. 
+            ///   The token can be used to pin the lock permanently through the Deferral API and, with that, take the message out of the 
+            ///   regular delivery state flow. This property is read-only.
```
nemakam: I'm not sure I get this. Deferral and receiving deferred message happens through sequence numbers and not lockToken. So what do you mean by "pinning the lock permanently"?

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -328,7 +394,13 @@ public DateTime LockedUntilUtc
                 internal set => this.lockedUntilUtc = value;
             }
 
-            /// <summary>Gets the unique number assigned to a message by Service Bus, for this entity.</summary>
+            /// <summary>Gets the unique number assigned to a message by Service Bus.</summary>
+            /// <remarks>
+            ///     The sequence number is a unique 64-bit integer assigned to a message as it is accepted 
+            ///     and stored by the broker and functions as its true identifier. For partitioned entities, 
+            ///     the topmost 16 bits reflect the partition identifier. Sequence numbers monotonically increase 
+            ///     and are gapless. They roll over to 0 when the 48-64 bit range is exhausted. This property is read-only.
```
nemakam: >and are gapless. [](start = 20, length = 16)

There can be perceived gaps in the case of say ScheduledMessages. Though there won't be gaps on the server side, the fact might not hold true for the receivers. Should we remove this?

clemensv: Scheduled messages get sequenced as they get put into the queue. How is there a gap?

src/Microsoft.Azure.ServiceBus/Message.cs
clemensv: You get a message under a lock and deferral conceptually allows you to hold on to that lock permanently. Defer acts on the lock.

src/Microsoft.Azure.ServiceBus/Message.cs
clemensv: Since they're set by the system that appears to be implied for a fresh message.

src/Microsoft.Azure.ServiceBus/Message.cs
clemensv: We don't need all info in all places.

src/Microsoft.Azure.ServiceBus/Message.cs
clemensv: I'm adding a link here https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messages-payloads to the class description.

src/Microsoft.Azure.ServiceBus/Message.cs
nemakam: @clemensv, Lets say we schedule a message. It gets enqueued as sequence number 10. And then when the message is actually pushed to the active queue, we get the next active sequence number and stamp it (say 15). 
For a customer who doesn't know or care about the internals of broker, he would see a missing sequence number of 10. 
Similar things happens with Deduplication feature as well. The duplicated message gets its own sequence number, but is never inserted into the queue. So the customer will never see that sequence number.
Adding "gapless" to the contract enforces us to follow it and taking that contract out is a difficult challenge. I think we should just stick to "monotonically increasing numbers" and not commit ourselves to "gapless"

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: @clemensv 👆

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: I'd almost agree. Except [PartitionKey API documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.brokeredmessage.partitionkey?view=azureservicebus-4.1.1#Microsoft_ServiceBus_Messaging_BrokeredMessage_PartitionKey), like other API docs, should probably have that info.

src/Microsoft.Azure.ServiceBus/Message.cs
SeanFeldman: 👍 If may suggest to split `Property Name` column and `mapping property`

clemensv:
@nemakam removed the "gapless" wording.

PR #350, opened 2017-10-09
Using PublicApiGenerator nuget

### LINKED ISSUE #315: Move over to using the PublicApiGenerator nuget
do u want a PR?

SeanFeldman:
Fixes #315 

**On hold** until my PR to allow namespace white listing is merged.


test/Microsoft.Azure.ServiceBus.UnitTests/API/APIApprovals.cs
```diff
@@ -1,10 +1,14 @@
 ﻿#if NET461
+
+using System.IO;
+using System.Reflection;
```
nemakam: Move inside namespace?

SeanFeldman: Those are duplicates. I'll remove them.

SeanFeldman:
@nemakam PR is ready for review. With the new API Approver the unnecessary code is removed and ASB namespaces can be white listed. 

Please review

test/Microsoft.Azure.ServiceBus.UnitTests/Microsoft.Azure.ServiceBus.UnitTests.csproj
```diff
@@ -30,24 +30,9 @@
...
-      <Version>0.9.6.4</Version>
-    </PackageReference>
-    <PackageReference Include="PublicApiGenerator">
-      <Version>6.0.0</Version>
-    </PackageReference>
-    <PackageReference Include="WindowsAzure.ServiceBus">
-      <Version>4.1.2</Version>
-    </PackageReference>
+    <PackageReference Include="ApprovalTests" Version="3.0.13" />
+    <PackageReference Include="PublicApiGenerator" Version="6.*" />
```
nemakam: Shouldnt the version be minimum of 6.1.0?

SeanFeldman: It will take the latest minor. PublicApiGenerator follows SemVer, so I figured it would be ok to lock major only and leave the rest dynamic. This way we don't have to update for every minor/patch.
Any objections?

nemakam: So what happens when you already have say 6.0.0 in your local cache? Will it still try to fetch the latest minor (6.1.0) or use the local cache?

SeanFeldman: It will retrieve 6.1.0 and store that in your cache.

nemakam: That means, on every 'build', it will try to not only look at local cache, it will try to make a network call to find the latest version of that nuget?
Rather we should just use `[6.1.0,*)` 
(Not sure what the exact syntax is, but you get the point)

SeanFeldman: `="6.4.0"` (its's the latest). I'll change it to make things simpler if you'd like.

nemakam:
Thanks @SeanFeldman (y)

PR #331, opened 2017-09-19
Updating version to 2.0

nemakam:
1. Updating client version to 2.0 since we have a breaking change (Interface changes in `IMessageReceiver`)
2. Bumping up dotnet framework version to netstandard 2.0. net 46 is required for functioning of web sockets

PR #353, opened 2017-10-11
Reverting interface changes to push 1.1.0 

nemakam:
Reverting #331 #309 #334 and #341 which change the interface. 
The idea is to release 1.1.0 with all the bug fixes and then work on 2.0 with all the interface changes as a separate release. Customers who do not wish to have breaking changes but still are blocked on the bug fixes can use 1.1.0 and customers waiting for the updated APIs can use 2.0.


SeanFeldman:
@nemakam did you intend to close this? If yes, you might want to delete the `revertRewnew` branch.

Can I suggest to take a different strategy for support versions?

For v1 support, have a `support-1.0` branch and do the work on that one. Whenever a new major version X is about to be released, create a `support-x.0` so that you have it stored and then you won't need to revert anything. Also, allows to backport changes in an easier manner.

PR #362, opened 2017-10-25
Fix #361

### LINKED ISSUE #361: Message Pump erroneously reports out expected ObjectDisposedException when the pump is shut down
https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs#L71

## Actual Behavior
1. MessageReceiver with registered callback is closed 
2. Exception callback is called with ObjectDisposedException when receiver is closed

## Expected Behavior
1. Message receiver with registered callback is closed
2. Exception callback is not called since the exception is an expected side effect of tearing down the receiver 

## Versions  
- OS platform and version: all
- .NET Version: all
- NuGet package version or commit ID: all

clemensv:
Fixes #361 

Suppresses output of ObjectDisposedException as the receive loop shuts down

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
+++ b/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs
@@ -68,8 +68,12 @@ namespace Microsoft.Azure.ServiceBus
                 }
                 catch (Exception exception)
                 {
-                    MessagingEventSource.Log.MessageReceivePumpTaskException(this.messageReceiver.ClientId, string.Empty, exception);
-                    await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.Receive).ConfigureAwait(false);
+                    // We will not report an ObjectDisposedException bubbling up as we're stopping the pump
+                    if (!(exception is ObjectDisposedException && this.pumpCancellationToken.IsCancellationRequested))
+                    {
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@clemensv the same issue can be observed with WindowsAzure.ServiceBus client as well.
Is this fix going to be ported as well? Thanks.

vinaysurya:
>                     if (!(exception is TaskCanceledException))

Can you please add that check here as well.

---
Refers to: src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs:213 in bfe8c9e. [](commit_id = bfe8c9e056198f5335c3ea0293b411b42cf6b9aa, deletion_comment = False)

vinaysurya:
Also add the same checks to the SessionReceivePump as well.

clemensv:
@SeanFeldman These are different projects in different places in their lifecycle. Forward progress is here. I don't see us backporting issues into the NETFX client that aren't security/reliability issues and for which we don't have active support cases or where there is no good workaround. For this case, you can know that you just closed the client, and you can then suppress work inside the callback; customers who need to do that in NETFX likely already do.

SeanFeldman:
@vinaysurya @nemakam out of curiosity, PR was approved, but never merged. Any reason for hold off? Thanks.

vinaysurya:
Fixed with pr https://github.com/Azure/azure-service-bus-dotnet/pull/390

PR #370, opened 2017-11-01
Instruments operations on ServiceBus with DiagnosticsSource and introduces correlation between producer and consumer

lmolkova:
This change adds instrumentation with `DiagnosticSource` for public ServiceBus APIs:
- `MessageSender` (i.e. `QueueClient` and `TopicClient`): `SendAsync`, `ScueduleMessage` and `Cancel`
- `MessageReceiver`: `Receive`, `ReceiveDeffered`, `Peek`, `Abandon`, `Defer`, `Complete`, `DeadLetter`, `RenewLock`
- `MessageSession`: `AcceptMessageSession`
- `SubscriptionClient` : `AddRule`, `RemoveRule`, `GetRules`
- Message and session handlers

Tracing system (e.g. ApplicationInsights) may subscribe to ServiceBus DiagnosticSource and receive events about these operations including tracing context needed to correlate events and log any important information.

In absence of tracing system, diagnostics is disabled and performance cost of having it is ~zero.

We also introduce here telemetry correlation between producer and consumer: when diagnostics is enabled on a consumer, tracing context is injected into the message. When a producer receives such message, it extracts and restores context making it available for tracing system.

This also adds `Message.ExtractActivity()` public extension method that allows extracting the context from a message that is useful when handler API is NOT used to process messages.

Note: this change introduces 'standard' fields to pass tracing context through the queues to correlate telemetry:
`Diagnostic-Id` - uniquely identifies operation that sent message(s) to the queue
`Correlation-Context` - optional extended context (empty by default)

This is pretty similar to [HTTP Correlation protocol](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md) implemented in .NET HTTP stack.

Users are encouraged to use these properties on all platforms, not only .NET. 

As a result of this change, a tracing system may enable diagnostics information and receive events it's interested in making sure tracing context is propagated through the service bus, without ANY changes in user code (except for non-handler processing).

Test coverage for the changes: 62.05 %

-------

Instrumentation approach backgound:
-  [Activity UserGuide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md)
- [DiagnosticSource UserGuide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md)
- [.NET HTTP Client instrumentation](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/HttpDiagnosticsGuide.md)
- [HTTP Correlation protocol](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md)
- [EventHub .NET Client instrumentation (WIP)](https://github.com/Azure/azure-event-hubs-dotnet/commit/8f934200e93810bbd6fe76b647ed2e5073085cce)

@nemakam @TomMilos  @brahmnes

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
@@ -5,6 +5,7 @@ namespace Microsoft.Azure.ServiceBus.Core
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -48,6 +49,7 @@ namespace Microsoft.Azure.ServiceBus.Core
         readonly object messageReceivePumpSyncLock;
         readonly bool ownsConnection;
         readonly ActiveClientLinkManager clientLinkManager;
+        readonly ServiceBusDiagnosticSource diagnosticSource;
 
         int prefetchCount;
         long lastPeekedSequenceNumber;
@@ -132,7 +134,7 @@ namespace Microsoft.Azure.ServiceBus.Core
             this.PrefetchCount = prefetchCount;
             this.messageReceivePumpSyncLock = new object();
             this.clientLinkManager = new ActiveClientLinkManager(this.ClientId, this.CbsTokenProvider);
-
+            this.diagnosticSource = new ServiceBusDiagnosticSource(entityPath, serviceBusConnection.Endpoint);
             MessagingEventSource.Log.MessageReceiverCreateStop(serviceBusConnection.Endpoint.Authority, entityPath, this.ClientId);
         }
 
@@ -298,30 +300,46 @@ namespace Microsoft.Azure.ServiceBus.Core
 
             MessagingEventSource.Log.MessageReceiveStart(this.ClientId, maxMessageCount);
 
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.ReceiveStart(maxMessageCount) : null;
+            Task receiveTask = null;
+
             IList<Message> unprocessedMessageList = null;
+            IList<Message> processedMessageList = null;
             try
             {
-                await this.RetryPolicy.RunOperation(
+                receiveTask = this.RetryPolicy.RunOperation(
                     async () =>
                     {
-                        unprocessedMessageList = await this.OnReceiveAsync(maxMessageCount, operationTimeout).ConfigureAwait(false);
-                    }, operationTimeout)
-                    .ConfigureAwait(false);
+                        unprocessedMessageList = await this.OnReceiveAsync(maxMessageCount, operationTimeout)
+                            .ConfigureAwait(false);
+                    }, operationTimeout);
+                await receiveTask.ConfigureAwait(false);
+
+                MessagingEventSource.Log.MessageReceiveStop(this.ClientId, unprocessedMessageList?.Count ?? 0);
+
+                if (unprocessedMessageList == null)
+                {
+                    return unprocessedMessageList;
+                }
+
+                processedMessageList = await this.ProcessMessages(unprocessedMessageList).ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageReceiveException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
-
-            MessagingEventSource.Log.MessageReceiveStop(this.ClientId, unprocessedMessageList?.Count ?? 0);
-
-            if (unprocessedMessageList == null)
+            finally
             {
-                return unprocessedMessageList;
+                this.diagnosticSource.ReceiveStop(activity, maxMessageCount, receiveTask?.Status, processedMessageList);
             }
 
-            return await this.ProcessMessages(unprocessedMessageList).ConfigureAwait(false);
+            return processedMessageList;
         }
 
         /// <summary>
@@ -366,22 +384,33 @@ namespace Microsoft.Azure.ServiceBus.Core
 
             MessagingEventSource.Log.MessageReceiveDeferredMessageStart(this.ClientId, sequenceNumberList.Length, sequenceNumberList);
 
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.ReceiveDefferedStart(sequenceNumberList) : null;
+            Task receiveTask = null;
+
             IList<Message> messages = null;
             try
             {
-                await this.RetryPolicy.RunOperation(
+                receiveTask = this.RetryPolicy.RunOperation(
                     async () =>
                     {
                         messages = await this.OnReceiveDeferredMessageAsync(sequenceNumberList).ConfigureAwait(false);
-                    }, this.OperationTimeout)
-                    .ConfigureAwait(false);
+                    }, this.OperationTimeout);
+                await receiveTask.ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageReceiveDeferredMessageException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
-
+            finally
+            {
+                this.diagnosticSource.ReceiveDefferedStop(activity, sequenceNumberList, receiveTask?.Status, messages);
+            }
             MessagingEventSource.Log.MessageReceiveDeferredMessageStop(this.ClientId, messages?.Count ?? 0);
 
             return messages;
@@ -425,16 +454,29 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
 
             MessagingEventSource.Log.MessageCompleteStart(this.ClientId, lockTokenList.Count, lockTokenList);
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.CompleteStart(lockTokenList) : null;
+            Task completeTask = null;
 
             try
             {
-                await this.RetryPolicy.RunOperation(() => this.OnCompleteAsync(lockTokenList), this.OperationTimeout)
-                    .ConfigureAwait(false);
+                completeTask =
+                    this.RetryPolicy.RunOperation(() => this.OnCompleteAsync(lockTokenList), this.OperationTimeout);
+                await completeTask.ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageCompleteException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
+
                 throw;
+            }
+            finally
+            {
+                this.diagnosticSource.CompleteStop(activity, lockTokenList, completeTask?.Status);
             }
 
             MessagingEventSource.Log.MessageCompleteStop(this.ClientId);
@@ -456,16 +498,30 @@ namespace Microsoft.Azure.ServiceBus.Core
             this.ThrowIfNotPeekLockMode();
 
             MessagingEventSource.Log.MessageAbandonStart(this.ClientId, 1, lockToken);
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.DisposeStart("Abandon", lockToken) : null;
+            Task abandonTask = null;
+
             try
             {
-                await this.RetryPolicy.RunOperation(() => this.OnAbandonAsync(lockToken, propertiesToModify), this.OperationTimeout)
-                    .ConfigureAwait(false);
+                abandonTask = this.RetryPolicy.RunOperation(() => this.OnAbandonAsync(lockToken, propertiesToModify),
+                    this.OperationTimeout);
+                await abandonTask.ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageAbandonException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
+            finally
+            {
+                this.diagnosticSource.DisposeStop(activity, lockToken, abandonTask?.Status);
+            }
+
 
             MessagingEventSource.Log.MessageAbandonStop(this.ClientId);
         }
@@ -487,18 +543,29 @@ namespace Microsoft.Azure.ServiceBus.Core
             this.ThrowIfNotPeekLockMode();
 
             MessagingEventSource.Log.MessageDeferStart(this.ClientId, 1, lockToken);
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.DisposeStart("Defer", lockToken) : null;
+            Task deferTask = null;
 
             try
             {
-                await this.RetryPolicy.RunOperation(() => this.OnDeferAsync(lockToken, propertiesToModify), this.OperationTimeout)
-                    .ConfigureAwait(false);
+                deferTask = this.RetryPolicy.RunOperation(() => this.OnDeferAsync(lockToken, propertiesToModify),
+                    this.OperationTimeout);
+                await deferTask.ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageDeferException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
-
+            finally
+            {
+                this.diagnosticSource.DisposeStop(activity, lockToken, deferTask?.Status);
+            }
             MessagingEventSource.Log.MessageDeferStop(this.ClientId);
         }
 
@@ -520,18 +587,29 @@ namespace Microsoft.Azure.ServiceBus.Core
             this.ThrowIfNotPeekLockMode();
 
             MessagingEventSource.Log.MessageDeadLetterStart(this.ClientId, 1, lockToken);
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
+            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.DisposeStart("DeadLetter", lockToken) : null;
+            Task deadLetterTask = null;
 
             try
             {
-                await this.RetryPolicy.RunOperation(() => this.OnDeadLetterAsync(lockToken, propertiesToModify), this.OperationTimeout)
-                    .ConfigureAwait(false);
+                deadLetterTask = this.RetryPolicy.RunOperation(() => this.OnDeadLetterAsync(lockToken, propertiesToModify),
+                    this.OperationTimeout);
+                await deadLetterTask.ConfigureAwait(false);
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageDeadLetterException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
-
+            finally
+            {
+                this.diagnosticSource.DisposeStop(activity, lockToken, deadLetterTask?.Status);
+            }
             MessagingEventSource.Log.MessageDeadLetterStop(this.ClientId);
         }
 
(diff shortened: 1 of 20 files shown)


src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,771 @@
...
+
+    using Microsoft.Azure.ServiceBus.Diagnostics;
+
+    internal class ServiceBusDiagnosticSource
+    {
+        public const string DiagnosticListenerName = "Microsoft.Azure.ServiceBus";
+
+        public const string ExceptionEventName = "Microsoft.Azure.ServiceBus.Exception";
+        public const string ProcessActivityName = "Microsoft.Azure.ServiceBus.Process";
+        public const string ProcessActivityStartName = "Microsoft.Azure.ServiceBus.Process.Start";
```
SeanFeldman: `ProcessActivityStartName`, `ProcessSessionActivityName`, and `ProcessSessionActivityStartName` don't seem to be used.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,771 @@
...
+            return Start("Receive", () => new
+            {
+                MessageCount = messageCount,
+                Entity = this.entityPath,
+                Endpoint = this.endpoint
+            },
+            null);
+        }
+
+        internal void ReceiveStop(Activity activity, int messageCount, TaskStatus? status, IList<Message> messageList)
```
SeanFeldman: `messageCount` is problematic. It represents the maximum number of messages ASB client _might_ fetch, but not necessarily the actual number. The actul number of retrieved messages should be retrieved from `messageList.Length` to be accurate.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,771 @@
...
+                    Entity = this.entityPath,
+                    Endpoint = this.endpoint,
+                    Status = status ?? TaskStatus.Faulted
+                });
+            }
+        }
+
+        #endregion
+
+        internal void ReportException(Exception ex)
```
nemakam: >internal void ReportException(Exception ex) [](start = 8, length = 43)

Is there a paradigm of `ReportException` being associated with an `Activity`?

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: That's a good point. However image you are logging this event:
`Requested {messageCount}, received {messageList.Count} in a {Activity.Current.Duration}`.
I think it's useful to know how much did you request and how much did you receive in response.

As you notice, every 'activity' has a Start and Stop events and `messageCount` really makes sense in Start event payload ('Requesting {messageCount} messages'). However Start event is not really interesting, at least not as interesting as Stop event (when you know the result and duration). So we expect the majority of listeners to care about Stop events only (and reduce verbosity by disabling Start events), so we try to provide full context about every operation in Stop.

src/Microsoft.Azure.ServiceBus/MessageSession.cs
```diff
@@ -162,5 +166,71 @@ protected override void ThrowIfClosed()
...
+        {
+            Activity activity = this.diagnosticSource.GetSessionStateStart(this.SessionId);
+            Task<byte[]> getStateTask = null;
+            byte[] state = null;
+
+            try
+            {
+                getStateTask = this.OnGetStateAsync();
+                state = await getStateTask.ConfigureAwait(false);
+                return getStateTask.Result;
```
nemakam: >getStateTask.Result [](start = 23, length = 19)

return state?

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,771 @@
...
+        internal Activity AcceptMessageSessionStart(string sessionId)
+        {
+            return Start("AcceptMessageSession", () => new
+                {
+                    SessionId = sessionId,
+                    Entity = this.entityPath,
+                    Endpoint = this.endpoint
+                },
+                a => a.AddTag("SessionId", sessionId));
+        }
```
nemakam: How good are these with null params? Couple of methods here have good chance of being passed a null param (like this one). What happens in that case?

src/Microsoft.Azure.ServiceBus/SessionClient.cs
```diff
@@ -236,9 +245,18 @@ await this.RetryPolicy.RunOperation(() => session.GetSessionReceiverLinkAsync(se
...
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
+
                 await session.CloseAsync().ConfigureAwait(false);
                 throw AmqpExceptionHelper.GetClientException(exception);
             }
+            finally
+            {
+                this.diagnosticSource.AcceptMessageSessionStop(activity, sessionId, serverWaitTime, session, acceptMessageSessionTask?.Status);
```
nemakam: >session [](start = 100, length = 7)

we don't need to pass the `session` object. Don't think its very useful. 
If you are removing this, make sure instead of passing `sessionId`, pass `session.SessionId`

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: Nothing would break if there is a null value of sessionId, but I'd still prefer not to add null to Tags (just to optimize amount of data). 
I'll add null check then, thanks

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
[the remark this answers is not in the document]
nemakam: What I meant was, we are just reporting exception here without associating it with an activity. Shouldn't we be passing an activity?
We associate all starts and stops of operations across machines. Will we be able to associate the exception with a correlationId?

---
In reply to: [149178343](https://github.com/Azure/azure-service-bus-dotnet/pull/370#discussion_r149178343) [](ancestors = 149178343)

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -108,6 +111,7 @@ public SubscriptionClient(string connectionString, string topicPath, string subs
             this.ReceiveMode = receiveMode;
             this.TokenProvider = this.ServiceBusConnection.CreateTokenProvider();
             this.CbsTokenProvider = new TokenProviderAdapter(this.TokenProvider, serviceBusConnection.OperationTimeout);
+            this.diagnosticSource = new ServiceBusDiagnosticSource($"{topicPath}/{subscriptionName}", serviceBusConnection.Endpoint);
```
nemakam: >"{topicPath}/{subscriptionName}" [](start = 68, length = 32)

this should be replaced with `this.Path`

src/Microsoft.Azure.ServiceBus/SubscriptionClient.cs
```diff
@@ -444,15 +448,28 @@ public Task AddRuleAsync(string ruleName, Filter filter)
...
                 MessagingEventSource.Log.AddRuleException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
                 throw;
             }
+            finally
+            {
+                this.diagnosticSource.AddRuleStop(activity, description, addRuleTask?.Status);
```
nemakam: >this.diagnosticSource.AddRuleStop(activity, description, addRuleTask?.Status); [](start = 16, length = 78)

wrap inside isDiagnosticsEnabled?

nemakam: Similarly at other places

---
In reply to: [149190095](https://github.com/Azure/azure-service-bus-dotnet/pull/370#discussion_r149190095) [](ancestors = 149190095)

lmolkova: if IsDiagnosticsEnabled is false, activity is null, all Stop methods do nothing if activity is null

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -298,30 +300,46 @@ public Task<IList<Message>> ReceiveAsync(int maxMessageCount)
...
                 throw;
             }
-
-            MessagingEventSource.Log.MessageReceiveStop(this.ClientId, unprocessedMessageList?.Count ?? 0);
-
-            if (unprocessedMessageList == null)
+            finally
             {
-                return unprocessedMessageList;
+                this.diagnosticSource.ReceiveStop(activity, maxMessageCount, receiveTask?.Status, processedMessageList);
```
nemakam: >this.diagnosticSource.ReceiveStop(activity, maxMessageCount, receiveTask?.Status, processedMessageList); [](start = 16, length = 104)

Start-Stop tracking might end up being used as latency counters for simplicity purposes by customers. Lets move the `ReceiveStop` log before `await this.ProcessMessages(..)`

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1147,6 +1263,7 @@ static void CloseRequestResponseSession(RequestResponseAmqpLink requestResponseA
             return processedMessageList;
         }
 
+
```
nemakam: nit: remove

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -144,25 +147,39 @@ public Task SendAsync(Message message)
...
             }
+            finally
+            {
+                this.diagnosticSource.SendStop(activity, messageList, sendTask?.Status);
+            }
 
             MessagingEventSource.Log.MessageSendStop(this.ClientId);
         }
 
+
```
nemakam: nit: Lets not have double empty lines :)

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,773 @@
...
+                });
+            }
+        }
+
+        #endregion
+
+
+        #region ReceiveDeffered
+
+        internal Activity ReceiveDefferedStart(IEnumerable<long> sequenceNumbers)
```
pharring: Typo: Deferred

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
nemakam: @lmolkova  What I meant was, we are just reporting exception here without associating it with an activity. Shouldn't we be passing an activity?
We associate all starts and stops of operations across machines. Will we be able to associate the exception with a correlationId?

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: thanks, will fix!

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: Activity has static Current property that flows with the async calls, i.e. users can always use Activity.Current, it does not have to be explicitly passed to the event

lmolkova: e.g. in exception event handler you can write
```C#
Logger.LogError($"Exception {ex}, Id: {Activity.Current?.Id}")
```

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,773 @@
...
+        private readonly string entityPath;
+        private readonly Uri endpoint;
+
+        public ServiceBusDiagnosticSource(string entityPath, Uri endpoint)
+        {
+            this.entityPath = entityPath;
+            this.endpoint = endpoint;
+        }
+
+        public static bool IsEnabled()
```
pharring: Suggestion: Make this a property. i.e.
```C#
public static bool IsEnabled => DiagnosticListener.IsEnabled();
```

I think it reads more naturally at the call-site.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
SeanFeldman: Agreed. Therefore [`MessageCount`](https://github.com/Azure/azure-service-bus-dotnet/pull/370/files/8000966fc57dde2106488864075c512213abb084#diff-b0e75e823e0fccc7d74474f07ab71968R229) should not be assigned the value of `messageCount` but `messageList.Length` instead.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: it also creates a false impression on the caller side, that `ServiceBusDiagnosticSource.IsEnabled` value is static and could be called every time instead of caching the value (like it done now)

```C#
            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
            Activity activity = isDiagnosticsEnabled ? this.diagnosticSource.ReceiveDeferredStart(sequenceNumberList) : null;
            try
            {
                ...
            }
            catch (Exception exception)
            {
                if (isDiagnosticsEnabled)
                {
                    this.diagnosticSource.ReportException(exception);
                }
            }
```

with property it will be more natural to write
```C#
            Activity activity = ServiceBusDiagnosticSource.IsEnabled? this.diagnosticSource.ReceiveDeferredStart(sequenceNumberList) : null;
            try
            {
                ...
            }
            catch (Exception exception)
            {
                if (ServiceBusDiagnosticSource.IsEnabled)
                {
                    this.diagnosticSource.ReportException(exception);
                }
            }
```

While it's quite cheap to call `DiagnosticListener.IsEnabled()`,  I still prefer result to be cached for particular operations not only because of performance, but also consistency

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: I'm sorry, I did not understand it.
If I never received Start event (and this is a default case) and on Stop event I want to log 
`Requested {messageCount}, received {messageList.Count} in a {Activity.Current.Duration}.`
How would I do that if `messageCount` is `messageList.Count`? 

Would you be ok with renaming `MessageCount` to e.g. `RequestedMessageCount`?

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/SessionDiagnosticsTests.cs
```diff
@@ -0,0 +1,332 @@
...
+
+            Assert.True(events.IsEmpty);
+
+            // SessionPumpTaskAsync calls AcceptMessageSessionAsync() without cancellation token.
+            // Even after SessionPump is stopped, this Task may still wait for session during operation timeout
+            // It may interferee with other tests by acception it's sessions and throwing exceptions.
+            // So, let's wait for timeout and a bit more to make sure all created tasks are completed
+            sw.Stop();
+
+            await Task.Delay((int)(timeout - sw.Elapsed).TotalMilliseconds + 1000);
```
nemakam: > await Task.Delay((int)(timeout - sw.Elapsed).TotalMilliseconds + 1000); [](start = 11, length = 72)

Should have caught this before during the code review.
In every test we need to ensure we have a finally block which will do `await queueClient.CloseAsync()` and similarly for other clients

nemakam: Remove the `Task.Delay`. If the issue still persists, then its a client bug that we need to fix. Once `CloseAsync()` is invoked, it is expected that every thread it spawns is closed.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,773 @@
...
+
+            if (DiagnosticListener.IsEnabled(activityName, entityPath))
+            {
+                var tmpActivity = message.ExtractActivity(activityName);
+                setTags?.Invoke(tmpActivity);
+                
+                if (DiagnosticListener.IsEnabled(activityName, entityPath, tmpActivity))
+                {
+                    activity = tmpActivity;
+                    if (DiagnosticListener.IsEnabled(activityName + "Start"))
```
pharring: Should `"Start"` have a leading period?

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -298,21 +300,35 @@ public Task<IList<Message>> ReceiveAsync(int maxMessageCount)
 
             MessagingEventSource.Log.MessageReceiveStart(this.ClientId, maxMessageCount);
 
+            bool isDiagnosticsEnabled = ServiceBusDiagnosticSource.IsEnabled();
```
pharring: This is super nit-picky, I know: Consider either `areDiagnosticsEnabled` or `isDiagnosticsSourceEnabeld`. Why? Because "Diagnostics" are plural; whereas "DiagnosticsSource" is singular.

This might be a bit too contorted, but we might be able to avoid the problem altogether by eliminating the local `bool` and test for `activity != null` lower down.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -298,21 +300,35 @@ public Task<IList<Message>> ReceiveAsync(int maxMessageCount)
...
+                    }, operationTimeout);
+                await receiveTask.ConfigureAwait(false);
+
             }
             catch (Exception exception)
             {
                 MessagingEventSource.Log.MessageReceiveException(this.ClientId, exception);
+                if (isDiagnosticsEnabled)
+                {
+                    this.diagnosticSource.ReportException(exception);
```
pharring: Consider putting this before the call to `MessagingEventSource.Log.MessageReceiveException`. That way, I think, we will have a consistent nesting of EventSource (outer) and DiagnosticSource (inner) messages.

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/SessionDiagnosticsTests.cs
lmolkova: I close queueClient and other resources in Dispose method - executed after every xunit test.
It's easier than clean up in every test.

If I remove this, tests become unstable. I've tried to fix it with a small timeout, so `AcceptMessageSessionAsync` returns faster, but it seems it's not always enough and I've seen tests fail because of it.

I can create an issue for that and add comment that this workaround should be removed once the issue is fixed. In the meantime, tests will be stable. Is it ok?

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
lmolkova: right, nice catch!

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
lmolkova: Thanks, I'm renaming to isDiagnosticSourceEnabled.

Regarding check for `actvity != null`,  while it works for Stop event, for exceptions, we assume some users may want to have Exception events only. With this level of verbosity there will be no activities created.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
lmolkova: makes sence

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/SessionDiagnosticsTests.cs
lmolkova: created https://github.com/Azure/azure-service-bus-dotnet/issues/372

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -0,0 +1,773 @@
...
+            }
+
+            var sessionIds = messageList.Where(m => m.SessionId != null).Select(m => m.SessionId).Distinct().ToArray();
+            if (sessionIds.Any())
+            {
+                activity.AddTag(SessionIdTag, string.Join(",", sessionIds));
+            }
+        }
+
+        private void SetTags(Activity activity, Message message)
```
TomMilos: @lmolkova - shouldn't we set entity path and endpoint as tags here? these are very important for each monitoring solution so they should be made easy to consume

lmolkova: as we discussed, we cannot leverage fully 'generic listener' and have to implement the custom listener for ServiceBus. The custom listener will be parsing payloads and extracting endpoint and entity from them. 
I wanted to keep instrumentation side as lightweight as possible. Otherwise, I would have to come up with 'standard' tags (as I'm sure similar things exists in almost every other Azure service client) and we would have to stick with them forever on AI side.

Issue #373, opened 2017-11-08
Expose public constructors on exception classes, to allow mocks to throw exceptions


dshpak-iq:
## Actual Behavior
1. The exception classes `MessageLockLostException`, `MessagingEntityDisabledException`, `MessagingEntityNotFoundException`, `QuotaExceededException`, `ServiceBusTimeoutException`, and `SessionLockLostException` have no public constructors.
2. Because of this, mocks of interfaces such as `IQueueClient` cannot throw these exceptions, making it difficult to unit test code that interacts with this library.
3. Note that `ServiceBusCommunicationException` is both public and non-sealed.

## Expected Behavior
1. Exception classes have public constructors, so that the Service Bus library can be mocked out for unit-testing purposes.

## Versions  
- OS platform and version: Windows 10 Pro
- .NET Version: .NET Framework 4.6.1
- NuGet package version or commit ID: Microsoft.Azure.ServiceBus 2.0.0

dshpak-iq:
If the team agrees with this change in principle, I'm perfectly willing to submit a pull request to make this change.

nemakam:
I don't see any reason to keep it internal. 
@dshpak-iq , sure. Submit a PR and we can push it through..

PR #374, opened 2017-11-09
Expose public constructors on exception classes

dshpak-iq:
## Description

As discussed in https://github.com/Azure/azure-service-bus-dotnet/issues/373, this makes the constructors for exception classes public, so that when interfaces such as `IQueueClient` are mocked for unit testing, the mocks can throw the correct exceptions.

The API approval tests have been updated to reflect this change.


dshpak-iq:
@nemakam Sure thing, done!

nemakam:
Thanks for the contribution..

PR #422, opened 2018-02-15
Adding MessageSizeExceededException

### LINKED ISSUE #356: Implement MessageSizeExceededException instead of throw a NotImplementedException
## Actual Behavior
1. NotImplementedException

## Expected Behavior
1. MessageSizeExceededException

https://github.com/Azure/azure-service-bus-dotnet/blob/50b56c0961457a41a574ed062c805985147d15d2/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L398

Please reintroduce the 'MessageSizeExceededException', we've been using this to split the messages into smaller batches and now we have to catch a 'NotImplementedException'

## Versions  
- OS platform and version: All
- .NET Version: Core 2.0
- NuGet package version or commit ID: 2.0.0

nemakam:
Resolves #356

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
@@ -155,6 +155,11 @@ namespace Microsoft.Azure.ServiceBus.Amqp
                 return new QuotaExceededException(message);
             }
 
+            if (string.Equals(condition, AmqpErrorCode.MessageSizeExceeded.Value))
+            {
+                return new MessageSizeExceededException(message);
+            }
+
             return new ServiceBusException(true, message);
         }
 
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -466,10 +466,7 @@ namespace Microsoft.Azure.ServiceBus.Core
                         var size = (ulong)amqpMessage.SerializedMessageSize;
                         if (size > amqpLink.Settings.MaxMessageSize.Value)
                         {
-                            // TODO: Add MessageSizeExceededException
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.ServiceBus/Amqp/AmqpExceptionHelper.cs
```diff
@@ -155,6 +155,11 @@ static Exception ToMessagingContractException(string condition, string message,
                 return new QuotaExceededException(message);
             }
 
+            if (string.Equals(condition, AmqpErrorCode.MessageSizeExceeded.Value))
```
SeanFeldman: `MessageSender` will throw `MessageSizeExceededException` when a message larger than max size is attempted to be sent. What's the scenario here?

nemakam: This is still just a client. The service can also throw this exception. 
It should ideally not happen, but its possible.

SeanFeldman: You're saying a broker could throw an exception on a message large than the max size? Under what curcomstances that would happen?

nemakam: 1. No matter how many client validations exist, broker validation is a MUST (due to different clients that we have).
1. Before the message reaches broker, some more routing information is generally added in between. Ideally it should all be accounted for already by giving some buffer to the message size (between the first node which gets your message, and the actual destination broker machines). But lets say there are some issues that creep up, either caused by added data on gateway or due to different message size evaluations between client and broker, or due to wrong max message size sent on link, etc, broker can end up throwing this exception. 
1. As mentioned earlier, these things should NOT happen. But we have seen it happening, so its a good defensive code to have.

SeanFeldman: I did not think of option #2. 👍

PR #425, opened 2018-02-20
Message TTL always had a value set

### LINKED ISSUE #424: Message TimeToLive Not obeying Default
## Actual Behavior
If you set the `DefaultMessageTimeToLive` when creating a queue, sending messages to it does not obey that default value.  For example, we have a queue which we've set `DefaultMessageTimeToLive` to `TimeSpan.MaxValue` and we've inspected the queue description with ServiceBus Explorer and it is indeed set to MaxValue.  However, when we send messages to it via this library, they always get a TTL of Now+~3weeks.   Even if we explicitly set a value for `TimeToLive`, to some largish value, it gets misinterpreted to some small value.  It only works properly if you set it to small values, ie within days.  I have not yet found the threshold of where it starts causing values that are incorrect.

NOTE: This worked perfectly in the .NET Framework version of the ServiceBus Library. 

If you're wondering why we want a TTL of essentially forever, it's because this is a special queue for us, it stores our migration history for our various ServiceBus instances.

## Expected Behavior
1. The `DefaultMessageTimeToLive` should be obeyed if no `TimeToLive` is set on the Message.
2. The `TimeToLive` setting on a Message should work properly for values up to `TimeSpan.MaxValue`

## Versions  
- OS platform and version:  Windows 10 1709 x64 (with latest updates)
- .NET Version: 4.7.1
- NuGet package version or commit ID:  2.0

mscrivo:
This broke the ability to use the default TTL value on the underlying entity.

This fixes #424

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
             amqpMessage.Properties.GroupId = sbMessage.SessionId;
             amqpMessage.Properties.ReplyToGroupId = sbMessage.ReplyToSessionId;
 
-            if (sbMessage.TimeToLive != null)
+            if (sbMessage.TimeToLive != TimeSpan.MaxValue)
             {
                 amqpMessage.Header.Ttl = (uint)sbMessage.TimeToLive.TotalMilliseconds;
                 amqpMessage.Properties.CreationTime = DateTime.UtcNow;
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/AmqpConverterTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/AmqpConverterTests.cs
@@ -84,6 +84,7 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
             var replyTo = Guid.NewGuid().ToString();
             var replyToSessionId = Guid.NewGuid().ToString();
             var publisher = Guid.NewGuid().ToString();
+            var timeToLive = TimeSpan.FromDays(5);
 
             var sbMessage = new Message(messageBody)
             {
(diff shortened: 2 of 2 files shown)


nemakam:
>         public static AmqpMessage SBMessageToAmqpMessage(SBMessage sbMessage)

Could you add a test which takes a sbMessage without TTL set, and call this method and make sure the `amqpMessage.Header.Ttl` is not set.

---
Refers to: src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs:94 in 578d491. [](commit_id = 578d491ca5e5bc5409e4fe84802893c9ec5133b5, deletion_comment = False)

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -192,22 +196,17 @@ public DateTime ExpiresAtUtc
...
-            }
+            get => this.timeToLive;
 
             set
             {
-                TimeoutHelper.ThrowIfNonPositiveArgument(value);
-                this.timeToLive = value;
+                if (value != null)
+                {
+                    TimeoutHelper.ThrowIfNonPositiveArgument(value.Value);
```
SeanFeldman: Should there be another helper method on `TimeoutHelper` to take a nullable TimeSpan? It looks odd to test for `null` and yet call a guard method after.

src/Microsoft.Azure.ServiceBus/Message.cs
```diff
@@ -192,22 +196,17 @@ public DateTime ExpiresAtUtc
         ///      setting and it is silently adjusted if it does. 
         ///      See <a href="https://docs.microsoft.com/azure/service-bus-messaging/message-expiration">Expiration</a>
         /// </remarks>
-        public TimeSpan TimeToLive
+        public TimeSpan? TimeToLive
```
SeanFeldman: Strictly following SemVer, this would require a major version.

mscrivo:
I'll create a new PR with @nemakam's suggestion so we don't need to break API compatibility.

mscrivo:
@nemakam I've made the changes you requested.

nemakam:
@mscrivo Thanks for the PR. Appreciate it...

mscrivo:
My pleasure!  Would it be much trouble to get a pre release nuget package published?

nemakam:
@mscrivo  Released. https://github.com/Azure/azure-service-bus-dotnet/releases

mscrivo:
Thank you @nemakam, very much appreciated!

PR #429, opened 2018-03-11
expose the inner MessageReceiver  / MessageSender

pksorensen:
In our remakrs
<remarks>Use <see cref="MessageSender"/> or <see cref="MessageReceiver"/> for advanced set of functionality.

so these should be exposed ?

Otherwise there is no way to renew a message lock

## Description

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/QueueClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/QueueClient.cs
@@ -210,7 +210,7 @@ namespace Microsoft.Azure.ServiceBus
         /// </summary>
         public override IList<ServiceBusPlugin> RegisteredPlugins => this.InnerSender.RegisteredPlugins;
 
-        internal MessageSender InnerSender
+        public MessageSender InnerSender
         {
             get
             {
@@ -235,7 +235,7 @@ namespace Microsoft.Azure.ServiceBus
             }
         }
 
-        internal MessageReceiver InnerReceiver
+        public MessageReceiver InnerReceiver
         {
             get
             {


SeanFeldman:
@pksorensen this doesn't look right.

`QueueCliet` is intended to be used with message handler (using `RegisterMessageHandler`) where a auto-renew time can be defined. If you want to manually receive and extend lock, `MessageReceiver` is a client to use.

pksorensen:
Sorry - i read the remarks as they should be used and public.

I understand now that the message could also mean that I should use the clasess directly. Sorry about that.

I was migrating from WindowsAzure.ServiceBus package and there I could from CloudClient use the message pump and still also disable renewing the lock and do it manually.

I do find it odd that i can disable automatic renew of the lock but not renew it? If what you saying is right,  why would one want to disable auto renewing the lock?

SeanFeldman:
> I was migrating from WindowsAzure.ServiceBus package and there I could from CloudClient use the message pump and still also disable renewing the lock and do it manually.

I suggest raising specific question about that with clear code example (Gist or GitHub reference) of what you were doing and what's missing.

> I do find it odd that i can disable automatic renew of the lock but not renew it? 

Auto-renewal with the pump is intended to be used with a callback. Auto-renew time serves as an "extended lock duration" and because of that it would not allow never ending locking which is possible if you can renew endlessly.

> If what you saying is right, why would one want to disable auto renewing the lock?

Personally, I don't like it. It's a best-effort and not a guaranteed operation. Lock renewal is always caused by the client and _can_ fail. So if processing can take place under the MaxLockDuration of 5 minutes, rather use that. But that's me 🙂

nemakam:
@pksorensen 
Yes. The way to do it is use the classes directly :)

PR #434, opened 2018-03-22
Add a range to AMQP dependency package

### LINKED ISSUE #432: Exception when using Microsoft.Azure.Amqp 2.2
## Steps to reproduce
1. Install Microsoft.Azure.ServiceBus package. the package depends on  Microsoft.Azure.Amqp (>= 2.1.2), so version 2.1.2 is installed.
2. since the latest version of  Microsoft.Azure.Amqp is 2.2, upgrade it to this version
3. write code to send a message to a queue, and execute. I followed roughly [this tutorial](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dotnet-get-started-with-queues).
## Actual Behavior
An exception is thrown from the `queueClient = new QueueClient(ServiceBusConnectionString, QueueName);` line (in my case it was in a class library called from an Azure Function, but other than that it was essentially the same). Message text is: "Could not load file or assembly 'Microsoft.Azure.Amqp, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.":"Microsoft.Azure.Amqp, Version=2.1.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"
## Expected Behavior
Message should be sent without errors

## Versions  
- OS platform and version: Windows 10 Enterprise
- .NET Version: 4.6.1
- NuGet package version or commit ID: 2.0.0

SeanFeldman:
Fixes #432 (Microsoft.Azure.Amqp tight to a specific version).

I'm not sure about other dependecies, but those are also hardcoded to specific versions rather than a range. That doesn't look right IMO.

For example:
`Microsoft.Azure.Services.AppAuthentication` is using `1.1.0-preview`. The moment there's a preview2, this will be broken. Same with the rest. Thoughts?

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
+++ b/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
@@ -30,7 +30,7 @@
   </Target>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Amqp" Version="2.1.3" />
+    <PackageReference Include="Microsoft.Azure.Amqp" Version="[2.1.3, 3.0.0)" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.1.0-preview" />
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="3.17.2" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
--- a/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
+++ b/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
@@ -32,7 +32,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Amqp" Version="[2.1.3, 3.0.0)" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.1.0-preview" />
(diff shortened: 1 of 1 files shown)


nemakam:
Probably we could keep AppAuthentication as an exception as we are taking dependency on preview version and there is no GA version as of now. 
For others, do you think there's a problem with open ended range?

SeanFeldman:
Take for example `Microsoft.IdentityModel.Clients.ActiveDirectory`. It's set to `Version="3.17.2"` meaning anything >=3.17.2 Assuming the package is following SemVer, next major could introduce breaking changes. That would mean this library should verify version 4 of `Microsoft.IdentityModel.Clients.ActiveDirectory` is working prior to allowing it. For tha, a range would be required, ``Version="[3.17.2, 4.0.0)"`.

nemakam:
@SeanFeldman 
Could you go ahead make the range changes to all the references then?

Also, what are your opinions on how to track updates required to this? For e.g., If we put an upper limit of 4.0.0, then unless someone is watching over their team's updates, how would we track when a new version 4.0 is released? 
Or maybe do you think we should have an open ended range which will break the tests in AppVeyor?

SeanFeldman:
Tracking dependencies - ideal option is to automatically open an issue when a dependency change is detected. That would require automation.
A more primitive way - customers would raise an issue when a dependency major coming out. This will trigger notification for maintainers _and_ force to evaluate/test changes. It could be not just a new version, but also breaking changes in APIs.

nemakam:
@SeanFeldman 
Lets stick to the primitive way till we have a better way.
For now, could you please add the ranges to all the references?

SeanFeldman:
@nemakam done.
Also raised a placeholder issue for tracking dependency changes (#435).

PR #473, opened 2018-05-26
Fix infinite loop in ConcurrentExpiringSet and slow down cleanup to prevent a pegged CPU.

benmollet:
See issue for more details.

* Added 1000ms as the default delay between loops. Looking for guidance on if that's the right number.

* Also wasn't super sure on how best to test this so I tested the two external use cases and manually tested the exit strategy to prevent an infinite loop.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Primitives/ConcurrentExpiringSet.cs
+++ b/src/Microsoft.Azure.ServiceBus/Primitives/ConcurrentExpiringSet.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Azure.ServiceBus.Primitives
         {
             lock (this.cleanupSynObject)
             {
-                if (this.cleanupScheduled)
+                if (this.cleanupScheduled || this.dictionary.Count <= 0)
                 {
                     return;
                 }
@@ -51,6 +51,8 @@ namespace Microsoft.Azure.ServiceBus.Primitives
 
         async Task CollectExpiredEntries()
         {
+            await Task.Delay(TimeSpan.FromSeconds(30));
+
             lock (this.cleanupSynObject)
             {
                 this.cleanupScheduled = false;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.ServiceBus/Primitives/ConcurrentExpiringSet.cs
```diff
@@ -45,11 +45,11 @@ void ScheduleCleanup()
                 }
 
                 this.cleanupScheduled = true;
-                Task.Run(() => this.CollectExpiredEntries());
+                Task.Run(async () => await this.CollectExpiredEntries());
             }
```
nemakam: This wont completely help. Since every `AddOrUpdate()` call will end up calling `ScheduleCleanup`, and in a high throughput scenario, the moment one thread has finished setting `this.cleanupScheduled=false`, another thread might start working on `CollectExpiredEntries`. We didnt really end up waiting for the timespan.

I'd recommend moving 
```
if (this.dictionary.Count <= 0)
            {
                return;
            }
```
inside `ScheduleCleanup` just before we set `this.cleanupScheduled = true`.

And move `Task.Delay()` as the first line in `CollectExpiredEntries`

src/Microsoft.Azure.ServiceBus/Primitives/ConcurrentExpiringSet.cs
```diff
@@ -65,6 +65,13 @@ void CollectExpiredEntries()
                 }
             }
 
+            if (this.dictionary.Count <= 0)
+            {
+                return;
+            }
+
+            await Task.Delay(1000);
```
nemakam: >1000 [](start = 29, length = 4)

Represent this through `TimeSpan`. Its more readable.

nemakam: And 1000 is a small number. Maybe make it once in 30 seconds or so

---
In reply to: [191035182](https://github.com/Azure/azure-service-bus-dotnet/pull/473#discussion_r191035182) [](ancestors = 191035182)

src/Microsoft.Azure.ServiceBus/Primitives/ConcurrentExpiringSet.cs
benmollet: Makes sense, done

PR #476, opened 2018-05-30
Allow MessageSender to expose TransferDestinationPath it's using

SeanFeldman:
## Description

`Path` and `Connection` are already available. `TransferDestinationPath` is not.
W/o it, passing message senders around requires to also pass around transfer destination path, which is redundant considering that info can be made availabe as a read-only property.

## Legal

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/TransactionTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/TransactionTests.cs
@@ -240,6 +240,10 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                     ts.Dispose();
                 }
 
+                // Adding delay since transaction Commit/Rollback is an asynchronous operation.
+                // Operating on the same message should not be done.
+                await Task.Delay(TimeSpan.FromSeconds(2));
+
                 using (var ts = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
                 {
                     await receiver.CompleteAsync(deferredMessage.SystemProperties.LockToken);
@@ -285,6 +289,11 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                         async () => await sender.SendAsync(message2));
                     ts.Complete();
                 }
+
+                // Adding delay since transaction Commit/Rollback is an asynchronous operation.
+                // Operating on the same message should not be done.
+                await Task.Delay(TimeSpan.FromSeconds(2));
+
                 transaction.Rollback();
 
                 // Two complete operations to different partitions.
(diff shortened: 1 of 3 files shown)


nemakam:
@SeanFeldman Could you update the API list?

SeanFeldman:
@nemakam done. Love the fact we have an automated way to verify API changes 😉
If this could go in 3.0.2 milestone, I'd be very happy and remove unnecessary code in my implementation.

SeanFeldman:
@nemakam great, thanks. Will it be included in 3.0.2?

SeanFeldman:
Aside from that test (`TransactionThrowsWhenOperationsOfDifferentPartitionsAreInSameTransaction`), there's no other cases that would have rollback with completion right after. Or I don't understand what you want 😊

test/Microsoft.Azure.ServiceBus.UnitTests/TransactionTests.cs
```diff
@@ -285,6 +289,11 @@ public class TransactionTests
                         async () => await sender.SendAsync(message2));
                     ts.Complete();
                 }
+
+                // Adding delay since transaction Commit/Rollback is an asynchronous operation.
+                // Operating on the same message should not be done.
+                await Task.Delay(TimeSpan.FromSeconds(2));
```
nemakam: We'll need this delay after `transaction.Rollback()` :)

test/Microsoft.Azure.ServiceBus.UnitTests/TransactionTests.cs
```diff
@@ -306,6 +315,10 @@ public class TransactionTests
                     ts.Complete();
                 }
 
+                // Adding delay since transaction Commit/Rollback is an asynchronous operation.
+                // Operating on the same message should not be done.
+                await Task.Delay(TimeSpan.FromSeconds(2));
```
nemakam: Same as above

test/Microsoft.Azure.ServiceBus.UnitTests/TransactionTests.cs
```diff
@@ -240,6 +240,10 @@ public class TransactionTests
                     ts.Dispose();
                 }
 
+                // Adding delay since transaction Commit/Rollback is an asynchronous operation.
+                // Operating on the same message should not be done.
+                await Task.Delay(TimeSpan.FromSeconds(2));
```
nemakam: Similar fixes needed in
`TransactionalCompleteRollbackTest`
`TransactionRollbackWorksAcrossClientsUsingSameConnectionToSameEntity`

SeanFeldman: Added those. Might have gone a bit extra - please review 🙂

nemakam:
Few of them are good..
places where we do `send` -> `rollback` -> `receive` should be fine as different operations are not on the same message :)
But looks good.. Thanks for the changes :)

PR #485, opened 2018-06-14
Allow clients to report if they own or share the underlying connection string (Part 1)

### LINKED ISSUE #482: Impossible to know if underlying connection is shared or not
When clients are created, they can be instantiated with a `connection-string` or a `connection`.
The second method allows `connection` sharing. It is also exposed via [`IClientEntity.ServiceBusConnection`](https://github.com/Azure/azure-service-bus-dotnet/blob/5e5fd3333f0cc9759fa4f0c4a78379e6584d3bcb/src/Microsoft.Azure.ServiceBus/IClientEntity.cs#L53). Unfortunately, there's no was to know if `connection` is **owned** by a client or not. 

In scenarios, where a decision has to be made based on the fact that a client does or does not own a connection, there's no way to determine that information. The information _is_ stored internally on each client, and could be exposed via `IClientEntity` as a read-only property.

**Suggestion** - expose `OwnsConnection` read-only property on `IClientEntity` for all clients to implement.

Modifying `IClientEntity` would cause a breaking change and require a new major version.
Alternative route would be to introduce `OwnsConnection` in a minor release and expose as part of the 
`IClientEntity` interface in the next major.

## Versions  
- NuGet package version or commit ID: 3.0.2

SeanFeldman:
Fixes #482 

- Allow clients to report if they own or share the underlying connection string.
- Remove unnecessary `ownsConnection` field from all clients

This is part 1 out of 2 parts PRs. Follow up PR #486 will be completing this change in the next major release.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/ClientEntity.cs
+++ b/src/Microsoft.Azure.ServiceBus/ClientEntity.cs
@@ -55,6 +55,11 @@ namespace Microsoft.Azure.ServiceBus
         public abstract ServiceBusConnection ServiceBusConnection { get; }
 
         /// <summary>
+        /// Returns true if connection is owned and false if connection is shared.
+        /// </summary>
+        public bool OwnsConnection { get; internal set; }
+
+        /// <summary>
         /// Gets the name of the entity.
         /// </summary>
         public abstract string Path { get; }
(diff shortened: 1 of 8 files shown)


SeanFeldman:
@nemakam ready for review

SeanFeldman:
@nemakam please review

PR #490, opened 2018-06-15
Management client is not an entity client

SeanFeldman:
- Extract needed `EntityClient` behaviour into an internal client and encapsulate it.
- Add configureAwaits + remove async where's not needed

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/IManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/IManagementClient.cs
@@ -68,5 +68,7 @@ namespace Microsoft.Azure.ServiceBus.Management
         Task<bool> TopicExistsAsync(string topicName, CancellationToken cancellationToken = default);
 
         Task<bool> SubscriptionExistsAsync(string topicName, string subscriptionName, CancellationToken cancellationToken = default);
+
+        Task CloseAsync();
     }
 }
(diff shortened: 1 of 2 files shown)


SeanFeldman:
ManagementClient is a client, but why to polite the public API with what
clearly doesn't belong to it?

Take plugin related API. It has no place with management. Those methods
should not throw, but not be exposed at all.

Hope this clarifies the reason and intent.

Sean Feldman

SeanFeldman:
> Its going to be really dirty to encapsulate the ClientEntity 

Is this PR that dirty? 😃

In regards to plugins - we don't know what we don't know. Building what-if is not a good idea. The plugins where intended for incoming and outgoing message pipelines, which do not have equivalent with management operations.

At the end of the day, I'm not going to insist on this too much. Just ask not to dismiss this idea too fast.

nemakam:
> Is this PR that dirty? 😃

Haha.. Oops.. 

> In regards to plugins - we don't know what we don't know. Building what-if is not a good idea.

True in general. Not sure if plugins follow the same rule. Plugins are supposed to let users run wild with scenarios that the SDK developers are either not thinking of / working on.

> The plugins where intended for incoming and outgoing message pipelines

I definitely do not see why they are to be restricted? Think of a very very basic plugin - a logging plugin. It is supposed to work on any client / any API. 

Could you list all the reasons for you to propose this?

SeanFeldman:
1. Management client is about meta plane (entities management, not
messages). If it happens internally to require operations that somewhat
resemble with other clients, that should remain an internal detail of
implementation.
2. Plugins API are message pipelines specific. They should not be
re-purposed. Any other pipeline should have its own abstraction. Otherwise,
there's a risk of unavoidable reuse-abuse.***
3. Public API should only expose what it does, not what it could or cannot
with an exception.

* Recall the days Service Bus namespace was a kitchen sink just because
everything could fit in.

Boarding plane and hope to see you warming up towards my suggestion.

SeanFeldman:
Potentially to be replaced by #493

SeanFeldman:
Replaced by #493

PR #491, opened 2018-06-16
Performance tweaks

SeanFeldman:
- Make methods that don't have to be instance specific static
- Remove string comparisons for CRUD operations to determine port number


SeanFeldman:
@nemakam please review

PR #481, opened 2018-06-07
Management Operations

nemakam:
Introduces Serivce Bus Management operations - CRUD of entities
Fixes #65 

The API is exposed through `ManagementClient` which uses HTTP underneath to perform these operations. 
Summary of exposed operations:
1. `GetQueue` returns the `QueueDescription` which contains static information about the queue. 
1. `QueueDescription` is required to Create or Update the queue.
1. `ManagementClient.GetQueueRuntimeInfo` returns `QueueRuntimeInfo` which contains runtime information of the queue like the message count and the size.
1. `GetQueues` will return list of `QueueDescription` and will not contain runtime information.

Majority of the operations are similar to the ones exposed in older client (`Microsoft.ServiceBus` -> `NamespaceManager`), with following differences:
1. `GetQueue` now returns only static information as part of `QueueDescription` and runtime  information is part of `QueueRuntimeInfo`.
1. Default value of `DuplicateDetection` feature is 1 minute.

PR #493, opened 2018-06-18
Management client is not an entity client (take 2)

SeanFeldman:
My flight got cancelled and I had some spare time.

Took what I've started in #490 and re-viewed it.
The `ManagementClient` was using an internal client for no good reason. All it needed was a connection string to extract some info and that's it. Things such as retry policy, operation timeouts, and token provider are all defaults anyways. 

Furthermore, the dependency of  of all the descriptor files on the management client was wrong. They should not know anything about it. Parsing/serializing was moved into extensions.

Additionally, management client was doing what it shouldn't do - validation. There's already an entity name helper class that does the work and that abstraction works better.

Final result is:
- Management class that depends only on necessary extensions
- Validations are done in the appropriate helper
- All descriptors are serialization/parsing agnostic and are contain no reference to the management client

@nemakam please give a good review. I believe this could be a much better approach than #490 and the original code layout. Thanks.


SeanFeldman:
I’ll fix the tests and approve API once those are green.

SeanFeldman:
Given the changes to the public API, needs to be evaluated carefully if this is a suitable candidate for a minor or should it become a new major.

SeanFeldman:
GetRule(name) is throwing a `MessagingEntityNotFoundException`. Is that by design or should it return a `null` is rule is not found? Same applies to other get specific entity APIs.

SeanFeldman:
@nemakam rebased against your PR (#481)

nemakam:
@SeanFeldman 
Thanks.. Taking a look

src/Microsoft.Azure.ServiceBus/EntityNameHelper.cs
```diff
@@ -69,5 +72,64 @@ public static string FormatRulePath(string topicPath, string subscriptionName, s
         {
             return string.Concat(entityPath, PathDelimiter, TransferDeadLetterQueueName);
         }
+
+        // TODO: Should the specific check methods be made public for customers to use?
+        // That way customers don't have to re-implement this kind of checks in their own code
```
nemakam: That's an interesting thought. But I fear there would be lot more things that the service would check for validity which cannot be done on client side. There's going to be a fake sense of reliability when customers start using this helper which officially claims to `CheckValidQueueName`.

So unless customers request for it, IMO, we should not be making it public.

SeanFeldman:
I had a few questions @nemakam that are pending:
1. [Get specific entity API behaviour question](https://github.com/Azure/azure-service-bus-dotnet/pull/493#issuecomment-398140414)
1. [Exposing helper methods](https://github.com/SeanFeldman/azure-service-bus-dotnet/blob/adcb9eb403fdfc50916a3431d5af2fa6ff46d28e/src/Microsoft.Azure.ServiceBus/EntityNameHelper.cs#L76)
1. [Releasing minor/major because of breaking public API](https://github.com/Azure/azure-service-bus-dotnet/pull/493#issuecomment-398090677)

src/Microsoft.Azure.ServiceBus/Filters/FilterExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+                    return new TrueFilter();
+                case "FalseFilter":
+                    return new FalseFilter();
+                default:
+                    return null;
+            }
+        }
+
+        public static XElement Serialize(this Filter filter)
+        {
```
nemakam: FalseFilter and TrueFilter have been taken off. Merge issues?

SeanFeldman: Those two will fall under `SqlFilter` and that `Serialize` will be invoked as they both inherit from `SqlFilter`.

nemakam:
Regarding minor/major version, I don't see any changes in existing API other than `Equals()` method that has been overridden in each of the filter.
And personally, i don't even look at that as a breaking change. I'd rather look at it as a bug which has been fixed :) '
`3.1.0` version should be good. Thoughts?

---
In reply to: [398163969](https://github.com/Azure/azure-service-bus-dotnet/pull/493#issuecomment-398163969) [](ancestors = 398163969)

src/Microsoft.Azure.ServiceBus/Filters/FilterExtensions.cs
SeanFeldman: In that case, should handle it differently. Will update in a bit, leave this as TODO for now.

nemakam:
In the interest of not breaking my head going through each method that has been moved, could you please mention where all have you changed the `Serialize` or `Parse` methods?

SeanFeldman:
Those methods where moved to its appropreate extension class. E.g. `Parse` and `Serialize` for `QueueDescription` -> `QueueDescriptionExtensions`.

nemakam:
Yeah I realized they have been moved to extensions files. But for example in the case of filters, you had "cleaned up" few lines of code within `Serialize` method. 
Are there any other such places where the methods moved are not exactly same as the ones in the original file?

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
```diff
@@ -1,101 +1,95 @@
...
+            if (builder.SasKeyName != null || builder.SasKey != null)
+            {
+                return new SharedAccessSignatureTokenProvider(builder.SasKeyName, builder.SasKey);
+            }
 
-        public override void RegisterPlugin(ServiceBusPlugin serviceBusPlugin)
-        {
-            throw new NotImplementedException($"{nameof(ManagementClient)} doesn't support plugins");
+            // TODO: is this even possible?
+            throw new Exception("Could not create token provider from the provided connection string.");
```
nemakam: Yes. When they purposefully pass null when using ITokenProvider and don't have SAS information in the CsBuilder. 
Throw `ArgumentException` instead here.

SeanFeldman:
That was more of an exception than a rule as I didn't realize that True/False filters are re-using SqlFilter and the only difference was in seriazliation. BTW, I've fixed that in the latest commit.

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
SeanFeldman: `ArgumentException` should indicate a single parameter issue. In this case it's 2 things: 1) token provider is null _and_ connection string doesn't include what we need. I still think a general exception is OK, just with a different error message (feel free to suggest).

nemakam: Maybe remove the TODO then? :)

---
In reply to: [196201085](https://github.com/Azure/azure-service-bus-dotnet/pull/493#discussion_r196201085) [](ancestors = 196201085)

src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
```diff
@@ -176,186 +168,10 @@ public string ForwardDeadLetteredMessagesTo
...
-                    ));
-
-            return doc;
-        }
 
-        public bool Equals(QueueDescription other)
+        public bool Equals(QueueDescription otherDescription)
         {
-            if (this.Path.Equals(other.Path, StringComparison.OrdinalIgnoreCase)
+            if (otherDescription is QueueDescription other && this.Path.Equals(other.Path, StringComparison.OrdinalIgnoreCase)
```
nemakam: >(otherDescription is QueueDescription other [](start = 15, length = 43)

This is always true :)

SeanFeldman: PR was green on top of your branch 😉

SeanFeldman: > This is always true :)

Let me get to that in a separate issue 😃

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
SeanFeldman: Will be in PR #495

PR #497, opened 2018-06-19
Keep runtime infos simple

SeanFeldman:
Follow up on #493 
I've missed the `*-RuntimeInfo` part.


src/Microsoft.Azure.ServiceBus/Management/QueueRuntimeInfoExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    return ParseFromEntryElement(xDoc);
+                }
+            }
+
+            throw new MessagingEntityNotFoundException("Queue was not found");
+        }
+
+        // TODO: is this used?
+        static IList<QueueRuntimeInfo> ParseCollectionFromContent(string xml)
```
nemakam: Nope. Not used. We should remove that given that we don't support getting multiple runtime infos at same time

nemakam:
>         public MessageCountDetails MessageCountDetails { get; set; }

Meanwhile, this could be changed to internal as well

---
Refers to: src/Microsoft.Azure.ServiceBus/Management/SubscriptionRuntimeInfo.cs:19 in bc6bf16. [](commit_id = bc6bf1606d4d34be6352d9b018bcf37919f5c679, deletion_comment = False)

nemakam:
> Just on subscription or all 3 entity types?

`MesageCountDetails.set` is already internal on Q and Topic info. 
I'll change it in my code

PR #504, opened 2018-06-20
XML doco

SeanFeldman:
Targets PR #481
Depends on #503 

@nemakam is the approach to take? I'm basing of documentation on the world client as there're no tests for all exceptions.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/IManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/IManagementClient.cs
@@ -13,22 +13,7 @@ namespace Microsoft.Azure.ServiceBus.Management
     {
         // Create if not exist. Else throw.
 
-        /// <summary>
-        /// Creates a new queue in the service namespace with the given name.
-        /// </summary>
-        /// <remarks>Throws if a queue already exists.</remarks>
-        /// <param name="queueName">The name of the queue relative to the service namespace base address.</param>
-        /// <param name="cancellationToken"></param>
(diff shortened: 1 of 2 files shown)


nemakam:
The approach looks perfect. 👍  
Do you think its too much redundancy to mention same set of exceptions for each operation? Maybe we should just specify exceptions thrown in interface summary?

SeanFeldman:
Right now my biggest issue is that I don't know what specific exception can be thrown on the backend for each operation. I'm feeding off the namespace manager documentation atm, which is not ideal. Any recommendaitons how to handle that?

nemakam:
I look at it as all the exceptions thats been implemented in client and not in service. If client doesn't understand any exception, it anyway converts it into a generic `ServiceBusException`. So instead of focusing on server, how about just looking at client exception handler and document accordingly?
Checkout `ManagementClient.ValidateHttpResponse()` method. I think you already got almost everything right (with an exception of `QuotaExceededException`)

SeanFeldman:
Ok. Then a proper review would need to be conducted when I'm done.
I'll also do it on the `ManagementClient` and not the contract `IManagementClient`.

SeanFeldman:
@nemakam I'm going to pause this. Not sure how much of a value this is since you'll have to go over it and fix pretty much everything 🙁

nemakam:
@SeanFeldman 
Okay, I' merging this and will continue from where you left.

SeanFeldman:
Thanks @nemakam.

Probably would be good to align documentation with #67 (Path and Name confusion).

PR #513, opened 2018-06-22
Throwing when MessageSender is unable to successfully schedule messages

### LINKED ISSUE #456: MessageSender.OnScheduleMessageAsync returns 0 when AmqpResponseStatusCode is not OK
## Actual Behavior
Sometimes when `QueueClient.ScheduleMessageAsync(Message, DateTimeOffset)` is called, the returned sequence number is equal to 0, and the message is not added to the queue.

I had a look at the code and found that in method `MessageSender.OnScheduleMessageAsync(Message)`, if the status of the `AmqpResponseMessage` is not `AmqpResponseStatusCode.OK`, [a MessagingContractException is created, but never thrown](https://github.com/Azure/azure-service-bus-dotnet/blob/ebfe7eb051dfa1f98fb1e66d1f7d03b1c0b64910/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L619). As a result, the caller receives 0 as a sequence number and has to test this value in order to determine wether the scheduling succeeded or not.

## Expected Behavior
I suppose the MessagingContractException should be thrown when `AmqpResponseMessage.StatusCode` is not `AmqpResponseStatusCode.OK`.

Also, the current return statement `return sequenceNumbers?.FirstOrDefault() ?? 0;` makes me wonder if it's really possible for an `AmqpResponse` to contain zero sequence numbers (or several), when the operation is a `ScheduleMessageOperation` and it succeeded.

## Versions  
- OS platform and version: N/A.
- .NET Version: N/A.
- NuGet package version: Microsoft.Azure.ServiceBus 2.0.0.

nemakam:
Fixes #456

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -609,22 +609,26 @@ namespace Microsoft.Azure.ServiceBus.Core
                         {
                             entry[ManagementConstants.Properties.ViaPartitionKey] = message.ViaPartitionKey;
                         }
-                }
+                    }
 
                     request.Map[ManagementConstants.Properties.Messages] = new List<AmqpMap> { entry };
 
-                    IEnumerable<long> sequenceNumbers = null;
                     var response = await this.ExecuteRequestResponseAsync(request).ConfigureAwait(false);
                     if (response.StatusCode == AmqpResponseStatusCode.OK)
                     {
-                        sequenceNumbers = response.GetValue<long[]>(ManagementConstants.Properties.SequenceNumbers);
+                        var sequenceNumbers = response.GetValue<long[]>(ManagementConstants.Properties.SequenceNumbers);
+                        if (sequenceNumbers == null || sequenceNumbers.Length < 1)
+                        {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -609,22 +609,26 @@ static void CloseRequestResponseSession(RequestResponseAmqpLink requestResponseA
...
                     if (response.StatusCode == AmqpResponseStatusCode.OK)
                     {
-                        sequenceNumbers = response.GetValue<long[]>(ManagementConstants.Properties.SequenceNumbers);
+                        var sequenceNumbers = response.GetValue<long[]>(ManagementConstants.Properties.SequenceNumbers);
+                        if (sequenceNumbers == null || sequenceNumbers.Length < 1)
+                        {
+                            throw new ServiceBusException(true, "Could not schedule message successfully.");
+                        }
+
+                        return sequenceNumbers[0];
```
SeanFeldman: Is there a chance that there's be more than a single sequence number?

nemakam: No. Should not happen.

Issue #527, opened 2018-06-29
Closing MessageReciever does not cancel Recieve operation


SeanFeldman:
Separate issue stemmed from the conversation in https://github.com/Azure/azure-service-bus-dotnet/issues/439#issuecomment-400861621

```c#
var receiver = new MessageReceiver(connectionString, "queue", ReceiveMode.ReceiveAndDelete);

var receivingTask = receiver.ReceiveAsync(TimeSpan.FromMinutes(1));
await receiver.CloseAsync();

await receivingTask;
```

## Actual
Receive operation is not be terminated and `ObjecDisposedException` is thrown.

## Expected
Receive operation is cancelled and no exception is thrown

JohnLTaylor:
I just fixed a bug just like this in Azure/azure-event-hubs-dotnet and I see the same bug.  I will be submitting a PR soon

nemakam:
[Microsoft.Azure.Amqp 2.3.2](https://github.com/Azure/azure-amqp/releases/tag/v2.3.2) fixes this issue.

Issue #439, opened 2018-04-04
Cancellation support for queueClient.ReceiveAsync()


ig-sinicyn:
WindowsAzure.ServiceBus v4.1.8

**Use case:** I have a code that receives and processes messages in a loop, something like this: 
```cs
		BrokeredMessage msg;
		while ((msg = await queueClient.ReceiveAsync(timeout)) != null)
		{
			// message processing goes here
		}
```
Messages are rare, timeout set to 2-5 minutes.

I want to be able to cancel receive op immediately on app shutdown, but there's no way to do that. The `Receive()` method does not accept CancellationToken.
I cannot abort receive via disposing the queue client, too. The client may be used by another parts of code and I do not want to abort it in the middle of message processing.
Use small timeouts? Nope, 1-sec receive timeout will result in a millions of billed operations on service bus (we have a lot of instances that poll the queue).

**Suggestion:** add cancellation support for the Receive() calls, please:)

ig-sinicyn:
Ok, thanks!

SeanFeldman:
@nemakam I'm running into this issue right now with the new client.
`messageReceiver.Receive()` by default has a timeout that and there's no way to indicate to the receivers to cancel their wait. Ideally, it should be possible to pass a cancellation token into `Receive()` operation and just to cancel that cancellation token to stop receiver's receive execution.

nemakam:
@SeanFeldman  Doesn't calling `messageReceiver.CloseAsync()` stop the call?

SeanFeldman:
@nemakam no, it doesn't 🙁
It waits for the operation timeout and then throws `ObjectDisposedException, Cannot access a disposed object.Object name: 'FaultTolerantAmqpObject`1'.

<details><summary>Exception stacktrace</summary>
[...]

PR #528, opened 2018-06-30
Fix Bug #527 Closing MessageReciever does not cancel Recieve operation

JohnLTaylor:
[Fix Bug #527 Closing MessageReciever does not cancel Recieve operation](https://github.com/Azure/azure-service-bus-dotnet/issues/527).

OnCloseAsync will now wait for the MessageReceiverPump to terminate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -204,3 +204,6 @@ Buildx86retail.dat
 /build/OpenCoverReport/*
 /build/coverage.xml
 test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.received.txt
+
+#UsedForTestingPurposes
+nuget.config
--- a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
+++ b/src/Microsoft.Azure.ServiceBus/Amqp/AmqpConnectionHelper.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Azure.ServiceBus.Amqp
             var tcpTransportSettings = new TcpTransportSettings
             {
                 Host = networkHost,
-                Port = port < 0 ? AmqpConstants.DefaultSecurePort : port,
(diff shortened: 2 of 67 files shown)


SeanFeldman:
> OnCloseAsync will now wait for the MessageReceiverPump to terminate

The issue #527 states in expected behaviour:

> Receive operation is cancelled and no exception is thrown

`Close` operation should not wait for `Recieve()` to timeout.  It should close receiver's connection and cancel `Recieve()`. Also, note that the issue was raised for raw receive operation, the abstraction on top of it (message pump with callback).

JohnLTaylor:
There is a problem when the stateful service is changing from the active partition to non-active.  If the message pump task isn't complete before ClientEntity.CloseAsync completes any access to the reliable store will cause a FabricNotPrimaryException exception.

SeanFeldman:
@JohnLTaylor yes, Service Fabric (SF) will move around your services (stateful or stateless), closing them in the process. But that's a symptom, not the actual problem. The underlying problem is that currently ASB client does not stop receive operation if one is in progress and client is closed.

nemakam:
@JohnLTaylor 
This change might not be required. The aim is that `MessagePumpTaskAsync` gets out of loop when `CloseAsync()` is called. Most of the waiting methods have access to the `cancellationToken` and the `receiveAsync()` itself should ideally return the moment the connection is broken down.

JohnLTaylor:
@nemakam Without changing the interface, this is not possible as the cancellation token is not passed to messageReceiver.ReceiveAsync.

nemakam:
@JohnLTaylor Take a look at this discussion https://github.com/Azure/azure-service-bus-dotnet/issues/439#issuecomment-401473918.
Basically when you close the connection, all existing operations are expected to exit immediately. There are places when it doesn't. This is something that should be taken care by the underneath AMQP layer which manages all connections and messages through the networking channel. It will be fixed soon wihtout having to change the interface

nemakam:
Closing this as it as already fixed.

Issue #505, opened 2018-06-20
Enable GitHub WIP app


SeanFeldman:
PRs marked as [`WIP]` or `[do not merge]` should not be possible to merge.
This will prevent human errors.

PR #539, opened 2018-07-29
[WIP] Batching

### LINKED ISSUE #538: Support Batching
Allow building a batch of messages to be sent to ensure it doesn't exceed message maximum size and doesn't fail at the sending time.

Given that `Message.Size` is not accurate (#109), batching should handle size calculation and provide an API accumulate messages for sending in a safe manner.

SeanFeldman:
Fixes #538

The client allows to send a collection of messages (`SendAsync(IList<Message>)`) which is problematic when there's a large number of messages that would not fit into the maximum message size. As a result of that, an exception is thrown when the send operation is invoked.

This PR introduces a **safe** batching option. A `Batch` is constructed before send operation is invoked and messages are added using `bool TryAdd(Message)` API. As long as batch total size doesn't exceed the maximum, messages are added to it. Otherwise, message is not added and can be carried over to the next batch. With this API messages sent will not exceed maximum message size.

What's left:
- [x] Add tests to ensure common scenarios are verified (`BatchTests` and `SenderReceiverTests`)
- [x] Ensure received messages are not attempted to be sent using `Batch` (logic from `MessageSender.ValidateMessage()`)
- [x] Raise an issue for the next major to bring `SendAsync(Batch)` to `ISenderClient` --> #543
- [x] Log similar to other `SendAsync()` overloads
- [x] Retrieve message max size from the broker
- [x] Pass batched messages through plugins
- ~~**BLOCKED**~~\* Support Diagnostics similar to how `SendAsync(IList<Message>)` does it

\* blocked until Diagnostics can support a different approach of registering messages. See an experimental PR comments [here](https://github.com/SeanFeldman/azure-service-bus-dotnet/pull/1).

Opening this PR to discuss further implementation details

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
@@ -63,10 +63,16 @@ namespace Microsoft.Azure.ServiceBus.Core
 
         }
 
+        /// <summary>
+        /// Number of messages in batch.
+        /// </summary>
+        public int Length => datas.Count;
+
         private long Size => result.SerializedMessageSize;
 
+
         /// <summary>
-        /// Convert batch to AMQP message
+        /// Convert batch to AMQP message.
         /// </summary>
         /// <returns></returns>
         public AmqpMessage ToAmqpMessage()
--- a/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.ServiceBus.Core
     [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
     public class Batch : IDisposable
     {
-        private readonly long maximumBatchSize;
+        internal readonly ulong maximumBatchSize;
         private AmqpMessage firstMessage;
         private readonly List<Data> datas;
         private AmqpMessage result;
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.ServiceBus.Core
         /// Construct a new batch with a maximum batch size.
         /// </summary>
         /// <param name="maximumBatchSize">Maximum batch size allowed for batch.</param>
-        public Batch(long maximumBatchSize)
+        public Batch(ulong maximumBatchSize)
         {
             this.maximumBatchSize = maximumBatchSize;
             this.datas = new List<Data>();
@@ -68,7 +68,7 @@ namespace Microsoft.Azure.ServiceBus.Core
         /// </summary>
         public int Length => datas.Count;
 
-        internal long Size => result.SerializedMessageSize;
+        internal ulong Size => (ulong) result.SerializedMessageSize;
 
 
         /// <summary>
@@ -128,6 +128,6 @@ namespace Microsoft.Azure.ServiceBus.Core
             }
         }
 
-        private string DebuggerDisplay => $"Batch: size={Size} message count={datas.Count}";
+        private string DebuggerDisplay => $"Batch: size={Size} message count={datas.Count} maximum size={maximumBatchSize}";
     }
 }
--- a/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using Microsoft.Azure.Amqp;
 using Microsoft.Azure.Amqp.Framing;
 using Microsoft.Azure.ServiceBus.Amqp;
+using Microsoft.Azure.ServiceBus.Diagnostics;
 
 namespace Microsoft.Azure.ServiceBus.Core
 {
@@ -35,6 +36,8 @@ namespace Microsoft.Azure.ServiceBus.Core
         public bool TryAdd(Message message)
         {
             ThrowIfDisposed();
+
+            message.VerifyMessageIsNotPreviouslyReceived();
 
             var amqpMessage = AmqpMessageConverter.SBMessageToAmqpMessage(message);
 
--- a/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/Batch.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Threading.Tasks;
+
 namespace Microsoft.Azure.ServiceBus.Core
 {
     using System;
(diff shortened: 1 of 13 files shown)


SeanFeldman:
This PR could use #505 @nemakam 🙂

SeanFeldman:
@makam @vinaysurya I've got everything except diagnostics.
Diagnostics (except exceptions) cannot be integrated at the moment and would have to wait until an [alternative option](https://github.com/SeanFeldman/azure-service-bus-dotnet/pull/1) to implement it is introduced.

I'd like to start a review of this PR. The feature could be used with a caveat of not having full Diagnostics available if users require this feature until alternative diagnostics registration is available and added later.

SeanFeldman:
@nemakam @vinaysurya ping

SeanFeldman:
@nemakam @vinaysurya let me know when you're ready to start reviewing. Thanks.

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// <summary>
+        /// Construct a new batch with a maximum batch size and outgoing plugins callback.
+        /// <remarks>
+        /// To construct a batch at run-time, use <see cref="MessageSender"/>, <see cref="QueueClient"/>, or <see cref="TopicClient"/>.
+        /// Use this constructor for testing and custom implementations.
+        /// </remarks>
+        /// </summary>
+        /// <param name="maximumBatchSize">Maximum batch size allowed for batch.</param>
+        /// <param name="pluginsCallback">Plugins callback to invoke on outgoing messages regisered with batch.</param>
+        public Batch(ulong maximumBatchSize, Func<Message, Task<Message>> pluginsCallback)
```
nemakam: >public Batch [](start = 8, length = 12)

I think having this constructor will cause more confusion. We already have a way to create a batch on the sender. 

For testing as of now, we could make this internal. 
For customers who need to test this, they could mock a Sender (in the future version) and then create a batch accordingly. 

Having this constructor exposed like this will definitely create confusion to customers which is just explained in remarks (which is not the most accessible comment).

I would think especially the second param - callback is going to be even more confusing for a beginner user. 
Lets make this internal for now.

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
```
nemakam: nit: inside

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using Microsoft.Azure.Amqp;
+    using Microsoft.Azure.Amqp.Framing;
+    using Microsoft.Azure.ServiceBus.Amqp;
+    using Microsoft.Azure.ServiceBus.Diagnostics;
+
+    [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
+    public class Batch : IDisposable
```
nemakam: >Batch [](start = 17, length = 5)

Message - MessageBatch. 
Sounds more consistent. Maybe rename?

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+        public int Length => datas.Count;
+
+        internal ulong Size => (ulong) result.SerializedMessageSize;
+
+
+        /// <summary>
+        /// Convert batch to AMQP message.
+        /// </summary>
+        /// <returns></returns>
+        public AmqpMessage ToAmqpMessage()
```
nemakam: >public [](start = 8, length = 6)

This doesn't need to be public

src/Microsoft.Azure.ServiceBus/Extensions/MessageDiagnosticsExtensions.cs
```diff
@@ -149,5 +150,14 @@ internal static bool TryExtractContext(this Message message, out IList<KeyValueP
             }
             return false;
         }
+
+
+        internal static void VerifyMessageIsNotPreviouslyReceived(this Message message)
```
nemakam: >VerifyMessageIsNotPreviouslyReceived [](start = 29, length = 36)

This class is dedicated to diagnostics. Lets move this out of this class

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            if (datas.Count == 1)
+            {
+                firstMessage.Batchable = true;
+                return firstMessage;
+            }
+
+            if (originalMessageData.messageId != null)
+            {
+                result.Properties.MessageId = originalMessageData.messageId;
```
nemakam: >result.Properties.MessageId = originalMessageData.messageId; [](start = 16, length = 60)

What if `TryAdd` succeeded, and then you tried to copy over `MessageId` and other properties, and you crossed your size limit?

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            firstMessage = null;
+            result = null;
+        }
+
+        private void ThrowIfDisposed()
+        {
+            if (result == null)
+            {
+                throw new Exception("Batch is has been disposed and cannot be re-used.");
```
nemakam: >Exception [](start = 26, length = 9)

ObjectDisposedException

nemakam: >Batch is [](start = 37, length = 8)

typo

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
```diff
@@ -0,0 +1,144 @@
...
+            }
+        }
+
+        private string DebuggerDisplay => $"Batch: size={Size} message count={datas.Count} maximum size={maximumBatchSize}";
```
nemakam: >size={Size} message count= [](start = 51, length = 26)

Add a `.` or `;` or any identifier between two properties so that its easier to read

src/Microsoft.Azure.ServiceBus/Core/Batch.cs
SeanFeldman: Absolutely. It also removes the need to "remember" original message system properties and tuple that stores those. Nice!

SeanFeldman:
@nemakam integrated your feedback

SeanFeldman:
Back to green

src/Microsoft.Azure.ServiceBus/Core/MessageBatch.cs
```diff
@@ -0,0 +1,141 @@
...
+            return false;
+
+        }
+
+        /// <summary>
+        /// Number of messages in batch.
+        /// </summary>
+        public int Length => datas.Count;
+
+        internal ulong Size => (ulong) result.SerializedMessageSize;
```
SeanFeldman: @nemakam I'm contemplating to make this public.
Reason: to be able to know what batch size was sent.

SeanFeldman:

axisc:
@SeanFeldman - we are moving the underlying issue to the [Azure SDK for .NET] (https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/servicebus/Microsoft.Azure.ServiceBus) repo.

Once we archive this repo, the PR will be read only and we can leverage this test case in the new repo.

PR #546, opened 2018-08-06
Changing topicName param to topicPath in CreateRuleAsync, fixing variable typo in CreateRuleAsync

ryanspletzer:
topicPath is used elsewhere in other methods, this was the only one using topicName.

Also while looking at the code I discovered a typo where `EntityNameHelper.CheckValidSubscriptionName(topicName);` should have been `EntityNameHelper.CheckValidSubscriptionName(subscriptionName);` .

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
@@ -630,7 +630,7 @@ namespace Microsoft.Azure.ServiceBus.Management
         /// <summary>
         /// Adds a new rule to the subscription under given topic.
         /// </summary>
-        /// <param name="topicName">The topic name relative to the service namespace base address.</param>
+        /// <param name="topicPath">The topic name relative to the service namespace base address.</param>
         /// <param name="subscriptionName">The name of the subscription.</param>
         /// <param name="ruleDescription">A <see cref="RuleDescription"/> object describing the attributes with which the messages are matched and acted upon.</param>
         /// <param name="cancellationToken"></param>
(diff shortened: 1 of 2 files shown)


SeanFeldman:
There's still confusion around `path` vs `name`... see #67.
[According to the team](https://github.com/Azure/azure-service-bus-dotnet/issues/67#issuecomment-298722100), topic has a name, not a path 😀

ryanspletzer:
@nemakam PR has been re-opened. :)

nemakam:
@ryanspletzer 
COuld you fix the test - `ApproveAzureServiceBus`
There is a `ApiApprovals.ApproveAzureServiceBus.approved` file. You will need to update it with `ApiApprovals.ApproveAzureServiceBus.received.txt` file

ryanspletzer:
@nemakam Took me a sec to figure that out but it is done now. :)

ryanspletzer:
@nemakam Yep when I CTRL + S the approved.txt file in VS Code it auto-trimmed the trailing whitespace, which the unit test did not like, putting that back seems to have solved it.

SeanFeldman:
@ryanspletzer the file is auto-generated. If `.approved.txt` version is reformatted, then every time there's a change to the public API, `.received.txt` would need to be re-formatted as well. Probably better to keep it as-is and do what you've done 👍

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
```diff
@@ -634,7 +634,7 @@ public virtual async Task<SubscriptionDescription> CreateSubscriptionAsync(Subsc
         /// <summary>
         /// Adds a new rule to the subscription under given topic.
         /// </summary>
-        /// <param name="topicName">The topic name relative to the service namespace base address.</param>
+        /// <param name="topicPath">The topic name relative to the service namespace base address.</param>
```
SeanFeldman: Param name was changed, but not the description.

ryanspletzer: Good catch, fixed.

PR #561, opened 2018-08-21
Swithch to Microsoft SourceLink

SeanFeldman:
Microsoft now support SourceLinking. This PR switches project to use it.

Also, the package from now on will include PDB

image

and repository info

image

The latter will show up on NuGet.org.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
+++ b/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
@@ -32,7 +32,7 @@
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="[3.17.2, 4.0.0)" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="[5.2.2, 6.0.0)" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="None" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@vinaysurya @nemakam there's something flaky about the tests that are failing. This PR should not be failing based on the changes made. Unfortunately, I still cannot re-run the build as I don't have the option/privileges in AppVayor for this project.

image

src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
```diff
@@ -28,12 +31,7 @@
     <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="[3.17.2, 4.0.0)" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.4.1" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="[5.2.2, 6.0.0)" />
-  </ItemGroup>
-
-  <ItemGroup Label="SourceLink to embed PDBs with the assembly">
-    <PackageReference Include="SourceLink.Create.GitHub" Version="2.8.3" PrivateAssets="All" />
-    <PackageReference Include="SourceLink.Copy.PdbFiles" Version="2.8.3" PrivateAssets="All" />
-    <DotNetCliToolReference Include="dotnet-sourcelink-git" Version="2.8.3" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
```
nemakam: I think we may have an issue with releasing a stable version of our client having a beta dependency. Lets wait till the stable version is released.

SeanFeldman: @nemakam this is a tooling package that is not going to be included in the final result nuget package or pulled in by customers (`PrivateAssets="All"`). The tool is working and I've been using it to release to production already (not ASB client).

SeanFeldman: For the reference, see https://github.com/SeanFeldman/ServiceBus.AttachmentPlugin/blob/master/src/ServiceBus.AttachmentPlugin/ServiceBus.AttachmentPlugin.csproj#L25. Was used to release [3.0.0](https://www.nuget.org/packages/ServiceBus.AttachmentPlugin/3.0.0)

Issue #569, opened 2018-09-12
MessageSender abstraction is broken for Path when Send-Via is used


SeanFeldman:
## Actual Behavior

`MessageSender` supports [Send-Via feature](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-transactions#transfers-and-send-via). Send-Via is using a transfer queue to route message to their final destination when using transactional processing. The abstraction is represented by a public overload of `MessageSender` that takes in two paths 
1. `entitytPath`
1. `viaEntityPath`
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L127-L135

The abstraction has been copied from the old client and works, but it's broken when `Path` and `TransferDestinationPath` are queried back. That's because all public overloads are invoking an internal overload, swapping the two paths:
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L132

where internal overload is
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L137-L144

assigning `viaEntityPath` to `this.Path` and `entityPath` to `this.TransferDestinationPath`. Which is inverted.

## Expected Behavior
`Path` and Send-Via/`TransferDestinationPath` after MessageSender` constructor is invoked should be identical to the values passed in.

## Repro

LinqPad [repro script](https://github.com/Azure/azure-service-bus-dotnet/files/2373545/asb.MessageSender.Path.vs.TransferDestinationPath.bug.linq.txt) attached.

```c#
var sender1 = new MessageSender(connection, "path");
sender1.Path.Dump("Should be 'path'");
Assert.IsNull(sender1.TransferDestinationPath); // green

var sender2 = new MessageSender(connection, "path", "intermediate");
sender2.Path.Dump("Should be 'path'");
Assert.AreEqual("path", sender2.Path); // red, equals to "intermediate"
Assert.AreEqual("intermediate", sender2.TransferDestinationPath); // red, equals to "path"
```

## Recommendations

The fix should be released as a **major version** for the following reasons:
[...]

PR #570, opened 2018-09-12
[WIP] verification test for MessageSender abstraction is broken for Path when Send-Via is used

SeanFeldman:
**DO NOT MERGE**

Issue #569 

Adding verification tests expose the bug against `master`.

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/MessageSenderTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/MessageSenderTests.cs
@@ -1,0 +1,31 @@
+﻿namespace Microsoft.Azure.ServiceBus.UnitTests
+{
+    using Core;
+    using Xunit;
+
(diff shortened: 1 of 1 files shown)


axisc:
@SeanFeldman - we are moving the underlying issue to the [Azure SDK for .NET] (https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/servicebus/Microsoft.Azure.ServiceBus) repo.

Once we archive this repo, the PR will be read only and we can leverage this test case in the new repo.

SeanFeldman:
> we are moving the underlying issue to the Azure SDK for .NET repo.

Can I suggest to move and then comment for traceability? Right now, anyone looking at this issue would have no lead to the replacement issue in the new repo. Thank you.

SeanFeldman:
Ah, nevermind. The issue is moved, PR was not. All good @axisc

Issue #564, opened 2018-08-28
Since operator== is overloaded this null check should use object.ReferenceEquals


dlstucki:
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs#L299

This line is making a call to `public static bool operator ==(QueueDescription o1, QueueDescription o2)`. It doesn't result in stack overflow but does result in a scenario where `operator ==(QueueDescription o1, QueueDescription o2)` calls `Equals` which then calls `operator ==(QueueDescription o1, QueueDescription o2)`

Suggestion: use `if (object.ReferenceEquals(other, null))` instead.

nemakam:
Closed with #572

Issue #563, opened 2018-08-27
QueueDescription/TopicDescription are not forward compatible


dlstucki:
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Management/QueueDescriptionExtensions.cs#L13

If in the future some new property is added to QueueDescription/TopicDescription this existing code will drop that unknown new property upon reading the QueueDescription. Then if an update is performed the serialized version will not include the not understood property and it will clear that property instead of leaving it as it was.

DataContract solves the issue using IExtensibleDataObject. This current XElement based approach would need to store a collection of not understood properties when deserializing and serialize them on the other side.

nemakam:
Closed with #572

Issue #562, opened 2018-08-24
Possible null reference exception if response.Content == null


dlstucki:
This expression will return null instead of a Task if response.Content == null:
https://github.com/Azure/azure-service-bus-dotnet/blob/e278f87b0b3aa563ab055918e988baec9ec155a3/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs#L898

Calling await on a null Task results in a NullReferenceException.

nemakam:
Closed with #572

PR #572, opened 2018-09-21
Making Queue/Topic/Subscription descriptions forward compatible (#563) + Minor bugs

nemakam:
Fixes-
#563 - Making Queue/Topic/Subscription descriptions forward compatible by storing all unknown properties in a list and relaying it back during Update.
#564 - Equals() should not use == to check for null
#562 - NRE when content = null.
NRE when GetRules() is invoked on rules containing custom data types.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
@@ -899,7 +899,7 @@ namespace Microsoft.Azure.ServiceBus.Management
                 return null;
             }
 
-            var exceptionMessage = await response.Content?.ReadAsStringAsync();
+            var exceptionMessage = await (response.Content?.ReadAsStringAsync() ?? Task.FromResult(string.Empty));
             exceptionMessage = ParseDetailIfAvailable(exceptionMessage) ?? response.ReasonPhrase;
 
             if (response.StatusCode == HttpStatusCode.Unauthorized)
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
```diff
@@ -436,7 +438,9 @@ public virtual async Task<IList<SubscriptionDescription>> GetSubscriptionsAsync(
         /// <exception cref="ServerBusyException">The server is busy. You should wait before you retry the operation.</exception>
         /// <exception cref="ServiceBusException">An internal error or an unexpected exception occured.</exception>
         /// <remarks>You can simulate pages of list of entities by manipulating <paramref name="count"/> and <paramref name="skip"/>.
-        /// skip(0)+count(100) gives first 100 entities. skip(100)+count(100) gives the next 100 entities.</remarks>
+        /// skip(0)+count(100) gives first 100 entities. skip(100)+count(100) gives the next 100 entities.
```
SeanFeldman: ❤️ for adding remarks like this

src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
```diff
@@ -283,6 +284,12 @@ public string UserMetadata
             }
         }
 
+        /// <summary>
+        /// List of properties that were retrieved using GetQueue but is not understood by this version of client is stored here.
```
SeanFeldman: > List of properties that were retrieved using GetQueue but ~~is~~ **are** not understood by this version of client is stored here.

List is stored, properties are retrieved and are not understood, right?

src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
```diff
@@ -283,6 +284,12 @@ public string UserMetadata
...
+        /// These will be sent back to the service as-is when UpdateQueue is called on this QueueDescription.
```
SeanFeldman: >  on this QueueDescription.

Did you mean the list will be sent back when an already retrieved `QueueDescription` will be used in `UpdateQueue` call?

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
nemakam: :)

src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
nemakam: Fixed in #575

PR #580, opened 2018-10-03
Provide correct exception message

### LINKED ISSUE #579: MessageReceiver.CompleteAsync with empty list throws ArgumentNullException
## Actual Behavior
1. CompleteAsync(new List&lt;string&gt;())
2. throws ArgumentException with "Value cannot be null" ParamName=lockTokens

This is misleading because the parameter is NOT null. Why is it an error to supply an empty list? Just return without doing anything.

## Expected Behavior
1. CompleteAsync(new List&lt;string&gt;())
2. No exception thrown

## Versions  
- OS platform and version: Windows
- .NET Version: Standard 2.0
- NuGet package version: 3.1.0

SeanFeldman:
Fixes #579 
For an empty list do not throw argument null exception

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Resources.resx
+++ b/src/Microsoft.Azure.ServiceBus/Resources.resx
@@ -159,6 +159,9 @@
   <data name="TokenMissingExpiresOn" xml:space="preserve">
     <value>The provided token does not specify the 'ExpiresOn' value.</value>
   </data>
+  <data name="ListOfLockTokensCannotBeEmty" xml:space="preserve">
+    <value>List of lock tokens cannot be empty</value>
+  </data>
   <data name="NotSupportedPropertyType" xml:space="preserve">
     <value>'{0}' is not a supported type.</value>
   </data>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.ServiceBus/Resources.resx
```diff
@@ -159,6 +159,9 @@
   <data name="TokenMissingExpiresOn" xml:space="preserve">
     <value>The provided token does not specify the 'ExpiresOn' value.</value>
   </data>
+  <data name="ListOfLockTokensCannotBeEmty" xml:space="preserve">
```
axisc: nit - please fix typo ListOfLockTokensCannotBeEmpty in the next commit

SeanFeldman: @axisc I'm blind as  a 🦇 - what typo?

stjepan: ListOfLockTokensCannotBeEmty > ListOfLockTokensCannotBeEmpty

SeanFeldman: Thanks @stjepan. Blind like a bat I am 😂

SeanFeldman: PR #583 raised

PR #585, opened 2018-10-10
Added GetNamespaceInfo API

nemakam:
Added GetNamespaceInfo API which will provide information related to the namespace including the type of namespace and the tier/SKU (standard/premium), and MUs (if premium)

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
@@ -78,6 +78,18 @@ namespace Microsoft.Azure.ServiceBus.Management
             return clone;
         }
 
+        /// <summary>
+        /// Gets information related to the namespace.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns><see cref="NamespaceInfo"/> containing namespace information.</returns>
+        /// <remarks>Works with any claim (Send/Listen/Manage).</remarks>
+        public virtual async Task<NamespaceInfo> GetNamespaceInfoAsync(CancellationToken cancellationToken = default)
+        {
+            var content = await GetEntity("$namespaceinfo", null, false, cancellationToken).ConfigureAwait(false);
+            return NamespaceInfoExtensions.ParseFromContent(content);
+        }
+
         #region DeleteEntity
 
         /// <summary>
--- a/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.ServiceBus.Management
         }
 
         /// <summary>
-        /// Gets information related to the namespace.
+        /// Gets information related to the currently used namespace.
         /// </summary>
         /// <param name="cancellationToken"></param>
         /// <returns><see cref="NamespaceInfo"/> containing namespace information.</returns>
--- a/src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Management
+{
+    /// <summary>
+    /// Specifies the SKU/tier of the messaging namespace.
+    /// </summary>
+    public enum MessagingSku
+    {
+        /// <summary>
+        /// Basic namespace. Shared Resource. Only queues are available.
+        /// </summary>
+        Basic = 1,
+
+        /// <summary>
+        /// Standard namespace. Shared Resource. Queues and topics.
+        /// </summary>
+        Standard = 2,
+
+        /// <summary>
+        /// Premium namespace. Dedicated Resource. Queues and topics.
+        /// </summary>
+        Premium = 3
+    }
+}
--- a/src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
@@ -21,6 +21,11 @@ namespace Microsoft.Azure.ServiceBus.Management
         /// <summary>
         /// Premium namespace. Dedicated Resource. Queues and topics.
         /// </summary>
-        Premium = 3
+        Premium = 3,
+
+        /// <summary>
+        /// Other SKUs.
+        /// </summary>
+        Others = 99
     }
 }
--- a/src/Microsoft.Azure.ServiceBus/Management/NamespaceInfo.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/NamespaceInfo.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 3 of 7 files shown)


src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus.Management
+{
+    /// <summary>
+    /// Specifies the SKU/tier of the messaging namespace.
+    /// </summary>
+    public enum MessagingSku
```
SeanFeldman: `MessagingSku` --> `NamespaceSku`
Messaging SKU is vague. These are specific to ASB namespaces.

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Number of messaging units allocated for namespace. 
+        /// Valid only for <see cref="NamespaceType.Messaging"/> and <see cref="MessagingSku.Premium"/>
+        /// </summary>
+        public int MessagingUnits { get; set; }
+
+        /// <summary>
+        /// Alias for the namespace.
+        /// </summary>
+        public string Alias { get; set; }
```
SeanFeldman: Alias? That's new... is there a documentation on this?

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfoExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+                        nsInfo.Name = element.Value;
+                        break;
+                    case "NamespaceType":
+                        if (Enum.TryParse<NamespaceType>(element.Value, out var nsType))
+                        {
+                            nsInfo.NamespaceType = nsType;
+                        }
+                        else
+                        {
+                            nsInfo.NamespaceType = NamespaceType.Unknown;
```
SeanFeldman: Is this even possible?

src/Microsoft.Azure.ServiceBus/Management/NamespaceType.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Specifies the type of entities the namespace can contain.
+    /// </summary>
+    public enum NamespaceType
+    {
+        /// <summary>
+        /// Namespace contains messaging entities (queues, topics)
+        /// </summary>
+        Messaging = 0,
```
SeanFeldman: Technically, both ASB and Notification Hubs are messaging.
As a user, I'd affiliate more with `ServiceBus` namespace, contrasted with `Mixed`. `Mixed` would indicate Service Bus and Notification Hubs.

src/Microsoft.Azure.ServiceBus/Management/NamespaceType.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// Supported only for backward compatibility.
+        /// Namespace can contain mixture of messaging entities and notification hubs. 
+        /// </summary>
+        Mixed = 2,
+
+        /// <summary>
+        /// Unknown
+        /// </summary>
+        Unknown = 100,
```
SeanFeldman: How's that possible? What scenario is this?

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfoExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (nsInfoXml == null)
+            {
+                throw new ServiceBusException(true);
+            }
+
+            foreach (var element in nsInfoXml.Elements())
+            {
+                switch (element.Name.LocalName)
+                {
+                    case "CreatedTime":
```
nemakam: >case "CreatedTime": [](start = 20, length = 19)

case Alias

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfo.cs
SeanFeldman: Still think it's a bit vague

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfoExtensions.cs
SeanFeldman: Then it's not "Unknown", it's "Other" with an appropriate description

src/Microsoft.Azure.ServiceBus/Management/NamespaceType.cs
SeanFeldman: Then it's `Other` with some XML comments.

src/Microsoft.Azure.ServiceBus/Management/MessagingSku.cs
nemakam: By popular vote internally, I'm keeping this as-is. There is some reuse of code that we do and this sku enum is used across services. 

---
In reply to: [224154533](https://github.com/Azure/azure-service-bus-dotnet/pull/585#discussion_r224154533) [](ancestors = 224154533)

src/Microsoft.Azure.ServiceBus/Management/NamespaceInfo.cs
nemakam: For now I'm going to leave it as-is since multiple changes might come up for this. Once clear documentation is released, will update the xml text here.

---
In reply to: [224308639](https://github.com/Azure/azure-service-bus-dotnet/pull/585#discussion_r224308639) [](ancestors = 224308639)

src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs
```diff
@@ -78,6 +78,18 @@ public static HttpRequestMessage CloneRequest(HttpRequestMessage req)
             return clone;
         }
 
+        /// <summary>
+        /// Gets information related to the namespace.
```
SeanFeldman:
```suggestion
        /// Gets information related to the currently used namespace.
```

test/Microsoft.Azure.ServiceBus.UnitTests/Management/ManagementClientTests.cs
```diff
@@ -605,6 +605,14 @@ public async Task CorrelationFilterPropertiesTest()
             await client.DeleteTopicAsync(topicName);
         }
 
+        [Fact]
+        [DisplayTestMethodName]
+        public async Task GetNamespaceInfoTest()
+        {
+            var nsInfo = await client.GetNamespaceInfoAsync();
+            Assert.NotNull(nsInfo);
```
SeanFeldman: Should this be a test that asserts the retrieved information rather than just checking for null? I.e. `nsInfo.MessagingSku == MessagingSku.Standard`.
Now that not be the case if someone running it locally with a premium tier, but I'd not worry about that case too much. Most of the test executions are happening on CI where namespace type is always the same.

nemakam: I did think about that. I couldnt come to a conclusion so left it as-is. I can do that Assert

SeanFeldman: 👍 null could be misleading as the object would be there, but none of the right values would be in place 🙂

SeanFeldman:
@nemakam thank you for your contribution 😸

PR #590, opened 2018-10-18
Keep delivery count as-is when peeking a message

### LINKED ISSUE #589: Confusion about message's delivery count value when peeking messages
## Actual Behavior
1. Message delivery count is increased by 1

## Expected Behavior
1. Message delivery count should not increase 

This is more of a clarification question. Let's say I create a queue and send 1 message to that queue. Immediately after sending the message, I fetch that message using [`PeekAsync()`](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs#L804).

It is my understanding that because I am just peeking the message and not locking it, the delivery count of that message should not change and I am expecting the delivery count to be `0`. However what I am noticing is that the delivery count of that message is `1`. If we look at the code [`here`](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs#L232), you will notice that SDK is adding 1 to the value returned by the Service Bus.

I am curious to know why SDK is doing this.

SeanFeldman:
Fixes #589


SeanFeldman:
@nemakam ready to be reviewed

nemakam:
Could you add a new assert line for delivery count in one of the existing test for peek and peeklock?

SeanFeldman:
Added a test against real queue.

SeanFeldman:
@nemakam two tests running against real queues - one for Peek and one for Receive operations.

SeanFeldman:
@nemakam  in place

SeanFeldman:
Anything else that is missing on this PR @nemakam ?

SeanFeldman:
PR is completed and ready to be reviewed/merged

SeanFeldman:
Risking to piss off people here, but any chance to have this included in [3.2.0 release](https://github.com/Azure/azure-service-bus-dotnet/milestone/18)?

Issue #609, opened 2018-11-26
Provide a ValueTask return type for QueueClient.RegisterMessageHandler


bbougot:
Based on Stephen Toub's topic regarding ValueTask [here](https://blogs.msdn.microsoft.com/dotnet/2018/11/07/understanding-the-whys-whats-and-whens-of-valuetask/) it would be great considering adding/replacing the  QueueClient.RegisterMessageHandler return type as a ValueTask, which could vastly improve performances by reducing the amount of task allocations as the number of messages received grows.

SeanFeldman:
@bbougot are you referring to the registered handler `Func` using `RegisterMessageHandler(Func<Message, CancellationToken, Task> handler, ...)`?

(This will also require .NET Standard 2.1 or [System.Threading.Tasks.Extensions](https://www.nuget.org/packages/System.Threading.Tasks.Extensions) package.)

bbougot:
Yes absolutely.

SeanFeldman:
In that case it could be also applied on `exceptionReceivedHandler` (that is only awaited internally)` to be converted from `Func<ExceptionReceivedEventArgs, Task>` to `Func<ExceptionReceivedEventArgs, ValueTask>` as well.

SeanFeldman:
I've created a spike PR #611, but it really should be tested for performance improvements to assume using `ValueTask` will improve any performance. Would you like to take a stab at verifying performance impact @bbougot?

bbougot:
@SeanFeldman Bench setup detailed [here](https://github.com/Azure/azure-service-bus-dotnet/pull/611#issuecomment-442026452).

SeanFeldman:
Given the findings in the PR #611 and the fact that `ValueTask` should not be surfaced to the public API, I'm going to close this issue. Should you disagree with the decision @bbougot, please re-open and provide the reasons to continue on this route. Thanks.

PR #611, opened 2018-11-26
[SPIKE] Switch MessageHandlers to use ValueTask

SeanFeldman:
Experimental PR for issue #609 to see if switching MessageHandlers (message pumps) to `ValueTask` yields any performance benefits.

According to `ValueTask` [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netcore-2.1#remarks)

> the default choice for any asynchronous method should be to return a `Task` or `Task<TResult>`. **Only if performance analysis proves it worthwhile should a `ValueTask<TResult>` be used instead of a `Task<TResult>`**.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
@@ -45,23 +45,23 @@ namespace Microsoft.Azure.ServiceBus.Core
 
         /// <summary>
         /// Receive messages continuously from the entity. Registers a message handler and begins a new thread to receive messages.
-        /// This handler(<see cref="Func{Message, CancellationToken, Task}"/>) is awaited on every time a new message is received by the receiver.
+        /// This handler(<see cref="Func{Message, CancellationToken, ValueTask}"/>) is awaited on every time a new message is received by the receiver.
         /// </summary>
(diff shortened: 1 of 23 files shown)


bbougot:
I have bench dev branch and PR-611, I've extracted the two .diagsession that I link to this post:
https://www.dropbox.com/s/y35tanvdoybkbu3/dev-branch.diagsession?dl=0
https://www.dropbox.com/s/vqjezib4uart9s2/PR-611.diagsession?dl=0

I haven't time yet to compare both profiles but you should be able to do so.

I also made a project repro here: https://github.com/bbougot/PR611
You have only to create a Service Bus Queue and fill the two required values "QUEUE_NAME" and "Endpoint=" to the respective Azure Service Bus queue name and connection string.

Basically, it is made to send batch of 1000 messages every 100 milliseconds and by switching Task/ValueTask depending on the branch we are (dev/PR-611), you will be able to find the differences by profiling the memory using Visual Studio profiler.

SeanFeldman:
Doesn't seem like .diagsession are portable. I couldn't load those properly on two different computers.
I'll run the sample locally to profile it.

SeanFeldman:
@bbougot I played with the code a bit - don't really see a significant impact. Interestingly enough, ValueTask based version had slightly more Tasks allocations... Mind to share your results as text/screenshots?

SeanFeldman:
Closing as it doesn't seem to be worthy of proceeding.

PR #621, opened 2018-12-13
Do not retry when there's an ambient transaction

### LINKED ISSUE #615: RetryPolicy is causing transaction exception for connectivity issue
When an ambient transaction exists, operations such as `SendAsync` [enlist into existing ambient transaction](https://github.com/SeanFeldman/azure-service-bus-dotnet/blob/86cf1b11ff45e99ed204bd8aebea1d3bec46bcac/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L533-L537). `RetryPolicy` is taking that operation later and [executes](https://github.com/SeanFeldman/azure-service-bus-dotnet/blob/86cf1b11ff45e99ed204bd8aebea1d3bec46bcac/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs#L250). When an operation is failing to execute due to **connectivity problems**, `RetryPolicy` re-executes that operation, which includes repeated transaction enlistment. This in turn throws an exception `System.InvalidOperationException: 'Local transactions are not supported with other resource managers/DTC.'` thrown by [`AmqpTransactionManager.EnlistAsync()`](https://github.com/SeanFeldman/azure-service-bus-dotnet/blob/86cf1b11ff45e99ed204bd8aebea1d3bec46bcac/src/Microsoft.Azure.ServiceBus/Amqp/AmqpTransactionManager.cs#L40), completely masquerading connectivity problem.

<details>
<summary>Stacktrace</summary>
   at Microsoft.Azure.ServiceBus.Core.MessageSender.OnSendAsync(IList`1 messageList) in C:\source\azure-service-bus-dotnet\src\Microsoft.Azure.ServiceBus\Core\MessageSender.cs:line 562
   at Microsoft.Azure.ServiceBus.RetryPolicy.RunOperation(Func`1 operation, TimeSpan operationTimeout) in C:\source\azure-service-bus-dotnet\src\Microsoft.Azure.ServiceBus\RetryPolicy.cs:line 83
   at Microsoft.Azure.ServiceBus.RetryPolicy.RunOperation(Func`1 operation, TimeSpan operationTimeout) in C:\source\azure-service-bus-dotnet\src\Microsoft.Azure.ServiceBus\RetryPolicy.cs:line 105
   at Microsoft.Azure.ServiceBus.Core.MessageSender.SendAsync(IList`1 messageList) in C:\source\azure-service-bus-dotnet\src\Microsoft.Azure.ServiceBus\Core\MessageSender.cs:line 252
   at ReproApp.Program.Main(String[] args) in C:\Users\Sean\Desktop\RetryPolicyRepro\ReproApp\Program.cs:line 33
   at ReproApp.Program.<Main>(String[] args)
</details>

## Expected Behavior
1. Operation retried due to connectivity problem should not fail with incorrect transaction exception
1. Transaction for the same operation should not be enlisted more than once

## Steps to reproduce this issue:
1. Run the following code (bellow, available at this [repo](https://github.com/SeanFeldman/RetryPolicyRepro))
1. Block the outgoing TCP port 5671

```c#
static async Task Main(string[] args)
{
    var connectionString = Environment.GetEnvironmentVariable("AzureServiceBus_ConnectionString");

    var management = new ManagementClient(connectionString);

    if (!await management.QueueExistsAsync("queue"))
    {
        await management.CreateQueueAsync("queue");
    }

    var sender = new MessageSender(connectionString, "queue");

    while (true)
    {
        using (var tx = new TransactionScope(TransactionScopeOption.RequiresNew, TransactionScopeAsyncFlowOption.Enabled))
        {
            Debugger.Break();
            await sender.SendAsync(new Message(Encoding.Default.GetBytes(DateTime.Now.ToString("s"))));
            tx.Complete();
        }
    }

    await management.CloseAsync();
    await sender.CloseAsync();
}
```

## Versions  
- OS platform and version: Windows 10
- .NET Version: .NET Framework 4.6.1 / .NET Core 2.2
- NuGet package version or commit ID: 3.2.0

SeanFeldman:
Fixes #615 

Fix is following [recommendation](https://github.com/Azure/azure-service-bus-dotnet/issues/615#issuecomment-446329787) made on the issue:
> The RetryPolicy should understand whether the operation is within a transaction scope or not. If it is, it should never retry. And this should be documented properly.

## TODO
- [x] Implement change to skip retries when an ambient transaction is detected
- [x] Document this behavior (XMLdocs, anywhere else?)
- [x] Verify with tests
- [ ] Release as a hotfix 3.2.1

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
@@ -67,18 +67,27 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
             Log($"Sent {messageCount} messages");
         }
 
-        internal static async Task<IList<Message>> ReceiveMessagesAsync(IMessageReceiver messageReceiver, int messageCount)
+        internal static async Task<IList<Message>> ReceiveMessagesAsync(IMessageReceiver messageReceiver, int messageCount, TimeSpan timeout = default)
         {
             var receiveAttempts = 0;
             var messagesToReturn = new List<Message>();
+            var stopwatch = Stopwatch.StartNew();
 
-            while (receiveAttempts++ < TestConstants.MaxAttemptsCount && messagesToReturn.Count < messageCount)
+            if (timeout == default)
+            {
+                timeout = TimeSpan.Zero;
+            }
+
+            while (messagesToReturn.Count < messageCount && (receiveAttempts++ < TestConstants.MaxAttemptsCount || stopwatch.Elapsed < timeout))
             {
                 var messages = await messageReceiver.ReceiveAsync(messageCount - messagesToReturn.Count);
-                if (messages != null)
+                if (messages == null)
                 {
-                    messagesToReturn.AddRange(messages);
+                    await Task.Delay(TestConstants.WaitTimeBetweenAttempts);
+                    continue;
                 }
+
+                messagesToReturn.AddRange(messages);
             }
 
             VerifyUniqueMessages(messagesToReturn);
@@ -98,7 +107,7 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
                 var msg = await messageReceiver.ReceiveDeferredMessageAsync(sequenceNumber);
                 if (msg != null)
                 {
-                    messagesToReturn.Add(msg); 
+                    messagesToReturn.Add(msg);
                 }
             }
 
--- a/src/Microsoft.Azure.ServiceBus/RetryExponential.cs
+++ b/src/Microsoft.Azure.ServiceBus/RetryExponential.cs
@@ -8,7 +8,8 @@ namespace Microsoft.Azure.ServiceBus
 
     /// <summary>
     /// RetryPolicy implementation where the delay between retries will grow in a staggered exponential manner.
-    /// RetryIntervals will be computed using a retryFactor which is a function of deltaBackOff (MaximumBackoff - MinimumBackoff) and MaximumRetryCount
+    /// RetryIntervals will be computed using a retryFactor which is a function of deltaBackOff (MaximumBackoff - MinimumBackoff) and MaximumRetryCount.
+    /// <remarks>RetryPolicy will not be applied when an ambient transaction is found.</remarks>
     /// </summary>
     public sealed class RetryExponential : RetryPolicy
     {
--- a/src/Microsoft.Azure.ServiceBus/RetryPolicy.cs
+++ b/src/Microsoft.Azure.ServiceBus/RetryPolicy.cs
@@ -6,6 +6,7 @@ namespace Microsoft.Azure.ServiceBus
     using System;
     using System.Collections.Generic;
     using System.Threading.Tasks;
+    using System.Transactions;
     using Primitives;
 
     /// <summary>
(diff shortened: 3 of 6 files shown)


SeanFeldman:
ReceiveAndDelete tests are flaky imo.

E.g. `QueueClientTests.ReceiveDeleteTest` is failing. It's supposed to send 10 messages and receive back 10. The way it's implemented is using `TestUtility.ReceiveMessagesAsync(count)`. Internally, this method is running  a tight loop of 5 attempts to receive all messages from an entity. Problem with this approach is that tight loop might be too fast for the broker to return all messages, resulting in a flaky tests, failing due to incorrect number of messages returned.  For the test verification, I'll extend add a delay between attempts.

SeanFeldman:
Green and ready.

test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
```diff
@@ -67,18 +67,27 @@ internal static async Task SendMessagesAsync(IMessageSender messageSender, int m
...
+
+            while (messagesToReturn.Count < messageCount && (receiveAttempts++ < TestConstants.MaxAttemptsCount || stopwatch.Elapsed < timeout))
             {
                 var messages = await messageReceiver.ReceiveAsync(messageCount - messagesToReturn.Count);
-                if (messages != null)
+                if (messages == null)
                 {
-                    messagesToReturn.AddRange(messages);
+                    await Task.Delay(TestConstants.WaitTimeBetweenAttempts);
+                    continue;
```
nemakam: why this change?

test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
```diff
@@ -67,18 +67,27 @@ internal static async Task SendMessagesAsync(IMessageSender messageSender, int m
...
+            if (timeout == default)
+            {
+                timeout = TimeSpan.Zero;
+            }
+
+            while (messagesToReturn.Count < messageCount && (receiveAttempts++ < TestConstants.MaxAttemptsCount || stopwatch.Elapsed < timeout))
             {
                 var messages = await messageReceiver.ReceiveAsync(messageCount - messagesToReturn.Count);
-                if (messages != null)
+                if (messages == null)
```
nemakam: This line anyway goes for 60 seconds. Why introduce the stopwatch and timeout which is less than 60 seconds?

SeanFeldman: Despite all messages being sent to the broker and only after that attempted to be received, the broker returns a subset of messages and not all.  `ReceiveAsync(count)`  will not wait for 60 seconds, it will return immediately with whatever number of messages it can get. I think you're confusing this with `ReceiveAsync(count, timeout)` variant.

test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
SeanFeldman: Related to the [comment above](https://github.com/Azure/azure-service-bus-dotnet/pull/621/files#r241603056). Just burning through 5 attempts is not enough. The remainder of the messages can show up later. And there's no point to have a tight loop w/o some sort of backoff (liniar, 1 sec in this case).

test/Microsoft.Azure.ServiceBus.UnitTests/TestUtility.cs
SeanFeldman: BTW, I had to introduce this change after having random builds failing either on both targets (.NET Core and Framework) or one of those. Sometimes it would all pass. With this [fix](https://github.com/Azure/azure-service-bus-dotnet/pull/621#issuecomment-446853207), the test is [no longer flaky](https://ci.appveyor.com/project/vinaysurya/azure-service-bus-dotnet/builds/20975609) as it was.

nemakam: Yes my bad. I was thinking of something else.

SeanFeldman: All good. If I'd have to count number of times you've corrected me, I'd need more than two hands 😂

PR #624, opened 2018-12-17
[WIP] Specific optional SessionId option to listen to with SessionHandler

alberthoekstra:
This PR provides another setting on the ```SessionHandlerOptions``` class so you can specify a single SessionId when registering a SessionHandler.

Couldn't find the CONTRIBUTING.md file. Please link so I can check that out.

I've made this change based on [this](https://github.com/Azure/azure-service-bus/issues/131) issue.

I'll have to look into the test coverage for the changes.

# Extended general description
So basically what I needed was a handler that handles 1 specific session id. 

I made an application that does a request for data by sending a message to a specific topic.
The response will be sent with the ```ReplyToSessionId``` value from the original message into another queue (Response queue).

So the application that needs to receive the message just has to listen to messages in the response queue with its SessionId. Registering a session handler on a Queue client doesn't allow me to filter on a specific SessionId.

Thinking that it would be a waste of resources to just create an if statement in the message handler I found out it was easy to implement it into the project.

# Providing functionality with this PR
The best way to show what is possible with this PR is by code. 
First the current way of solving the problem:
```c#
 const string SessionId = "B1A405E2-C354-4F00-8E9E-8D71BA5C30B3";
var queueClient = new QueueClient(ConnectionString, QueueName);

queueClient.RegisterSessionHandler((session, message, arg3) =>
{
    if (!session.SessionId.Equals(SessionId, StringComparison.OrdinalIgnoreCase))
        return Task.CompletedTask;

    // process data here

    return Task.CompletedTask;
}, new SessionHandlerOptions(args =>
{
    // handle error here
    return Task.CompletedTask;
}));
```

New way of solving the problem:
```c#
const string SessionId = "B1A405E2-C354-4F00-8E9E-8D71BA5C30B3";
var queueClient = new QueueClient(ConnectionString, QueueName);

queueClient.RegisterSessionHandler((session, message, arg3) =>
{
    // process data here

    return Task.CompletedTask;
}, new SessionHandlerOptions(args =>
{
    // handle error here
    return Task.CompletedTask;
})
{
    SessionId = SessionId
});
```

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
@@ -88,7 +88,8 @@ namespace Microsoft.Azure.ServiceBus
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
-         /// <summary>Gets or sets the optional session ID.</summary>
+         /// <summary>Gets or sets a single session ID. This option is optional.
+         /// Note that the handler will only listen to this single session ID for its entire lifetime.</summary>
         public string SessionId { get; set; }
 
         internal bool AutoRenewLock => this.MaxAutoRenewDuration > TimeSpan.Zero;
--- a/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
@@ -88,8 +88,8 @@ namespace Microsoft.Azure.ServiceBus
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
-         /// <summary>Gets or sets a single session ID. This option is optional.
-         /// Note that the handler will only listen to this single session ID for its entire lifetime.</summary>
+         /// <summary>Gets or sets a single session ID. This option is optional.</summary>
+         /// <remark>Note that the handler will only listen to this single session ID for its entire lifetime.</remark>
         public string SessionId { get; set; }
 
         internal bool AutoRenewLock => this.MaxAutoRenewDuration > TimeSpan.Zero;
--- a/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
@@ -83,10 +83,13 @@ namespace Microsoft.Azure.ServiceBus
                 this.MaxConcurrentAcceptSessionCalls = Math.Min(value, 2 * Environment.ProcessorCount);
             }
         }
-
+        
         /// <summary>Gets or sets whether the autocomplete option of the session handler is enabled.</summary>
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
+
+        /// <summary>Gets or sets the optional session ID</summary>
+        public string SessionId { get; set; }
 
         internal bool AutoRenewLock => this.MaxAutoRenewDuration > TimeSpan.Zero;
 
--- a/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Azure.ServiceBus
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
-        /// <summary>Gets or sets the optional session ID</summary>
+         /// <summary>Gets or sets the optional session ID.</summary>
         public string SessionId { get; set; }
 
         internal bool AutoRenewLock => this.MaxAutoRenewDuration > TimeSpan.Zero;
--- a/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
@@ -88,10 +88,6 @@ namespace Microsoft.Azure.ServiceBus
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
-         /// <summary>Gets or sets a single session ID. This option is optional.</summary>
-         /// <remark>Note that the handler will only listen to this single session ID for its entire lifetime.</remark>
-        public string SessionId { get; set; }
-
         internal bool AutoRenewLock => this.MaxAutoRenewDuration > TimeSpan.Zero;
 
         internal int MaxConcurrentAcceptSessionCalls { get; set; }
--- a/src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
+++ b/src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
@@ -122,7 +122,13 @@ namespace Microsoft.Azure.ServiceBus
                     concurrentSessionSemaphoreAcquired = true;
 
                     await this.maxPendingAcceptSessionsSemaphoreSlim.WaitAsync(this.pumpCancellationToken).ConfigureAwait(false);
-                    var session = await this.client.AcceptMessageSessionAsync().ConfigureAwait(false);
+
+                    IMessageSession session;
+                    if(string.IsNullOrEmpty(sessionHandlerOptions.SessionId))
+                        session = await this.client.AcceptMessageSessionAsync().ConfigureAwait(false);
+                    else
+                        session = await this.client.AcceptMessageSessionAsync(sessionHandlerOptions.SessionId).ConfigureAwait(false);
+
                     if (session == null)
                     {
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -122,7 +122,13 @@ async Task SessionPumpTaskAsync()
                     concurrentSessionSemaphoreAcquired = true;
 
                     await this.maxPendingAcceptSessionsSemaphoreSlim.WaitAsync(this.pumpCancellationToken).ConfigureAwait(false);
-                    var session = await this.client.AcceptMessageSessionAsync().ConfigureAwait(false);
+
+                    IMessageSession session;
+                    if(string.IsNullOrEmpty(sessionHandlerOptions.SessionId))
+                        session = await this.client.AcceptMessageSessionAsync().ConfigureAwait(false);
+                    else
+                        session = await this.client.AcceptMessageSessionAsync(sessionHandlerOptions.SessionId).ConfigureAwait(false);
```
SeanFeldman: This means Session Handler will be pinned to the `SessionId` and process only one sessions for its entire lifetime. Is that the intent @alberthoekstra?

src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
```diff
@@ -83,11 +83,14 @@ public int MaxConcurrentSessions
                 this.MaxConcurrentAcceptSessionCalls = Math.Min(value, 2 * Environment.ProcessorCount);
             }
         }
-
+        
         /// <summary>Gets or sets whether the autocomplete option of the session handler is enabled.</summary>
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
+        /// <summary>Gets or sets the optional session ID</summary>
```
SeanFeldman: Assuming this PR goes through, I'd add a remark related to my comment about pinning to a single session.

SeanFeldman:
```suggestion
         /// <summary>Gets or sets the optional session ID.</summary>
```

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
```diff
@@ -122,7 +122,13 @@ async Task SessionPumpTaskAsync()
                     concurrentSessionSemaphoreAcquired = true;
 
                     await this.maxPendingAcceptSessionsSemaphoreSlim.WaitAsync(this.pumpCancellationToken).ConfigureAwait(false);
-                    var session = await this.client.AcceptMessageSessionAsync().ConfigureAwait(false);
+
+                    IMessageSession session;
+                    if(string.IsNullOrEmpty(sessionHandlerOptions.SessionId))
```
SeanFeldman: There's no need in `if/else`. `null` indicates any session id. Therefore, `sessionHandlerOptions.SessionId` can be passed for both cases.

SeanFeldman (commented):
Might want to add a test to verify the behavior.

@nemakam can you review this?

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
alberthoekstra: Allright, I’ll Change that tomorrow.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
alberthoekstra: Yes, that’s exactly what I want to happen.

alberthoekstra: Changed t. Check latest commit.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
alberthoekstra: Changed it. Check latest commit.

src/Microsoft.Azure.ServiceBus/SessionReceivePump.cs
SeanFeldman: In that case I'd suggest to elaborate in more details under the [general description](https://github.com/Azure/azure-service-bus-dotnet/pull/624#issue-239293202) to provide team with an idea what this PR is trying to accomplish and how it would work.

alberthoekstra: I’ve changed the description. Is this enough?

SeanFeldman (changes requested):
@alberthoekstra you'll need the following:
1. Update XML doco to indicate that only one session will be handled by message handler.
2. Update public API by running ApiApprovals tests and updating approved API to include changes to `SesssionHandlerOptions`

src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
alberthoekstra: I didn't know exactly what you meant. But I've extended the summary in the latest commit.

alberthoekstra:
@seanfeldman can you help me out with the ApiApprovals test? The file is not getting included into my project when I’m opening the solution in VS. 

It has something to do with the target framework but I can’t figure out what the problem is exactly.

SeanFeldman:
@alberthoekstra I can do that on your PR, but would need to have permissions to push to your repo.

SeanFeldman:
Note: once PR #625 is merged into `dev`, this branch would need to be rebased to include the changes.

src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
alberthoekstra: Changed it.

alberthoekstra:
PR #625 is merged.

I've added you as Collaborator in the fork repo.

SeanFeldman:
Public API changes approved [here](https://github.com/Azure/azure-service-bus-dotnet/pull/624/commits/b30d2f0b1cc4ae21f43e6b185d341dac84513e29)

SeanFeldman (dismissed):

SeanFeldman:
@alberthoekstra all looks good. 
This will be either released in the next minor or major, depending on the schedule.
Thank you.

src/Microsoft.Azure.ServiceBus/SessionHandlerOptions.cs
```diff
@@ -83,11 +83,15 @@ public int MaxConcurrentSessions
...
-
+        
         /// <summary>Gets or sets whether the autocomplete option of the session handler is enabled.</summary>
         /// <value>true if the autocomplete option of the session handler is enabled; otherwise, false.</value>
         public bool AutoComplete { get; set; }
 
+         /// <summary>Gets or sets a single session ID. This option is optional.</summary>
+         /// <remark>Note that the handler will only listen to this single session ID for its entire lifetime.</remark>
+        public string SessionId { get; set; }
+
```
nemakam: What happens when you set sessionId here and then set maxConcurrentSessions as well?
The other args do not hold up when you put a sessionId here.

SeanFeldman: Good question. I've missed that... normally there's be a default (2000) or user specified number of concurrent pumps. When `SessionId` is specified (i.e. not null), it should be short circuited to one pump only imo.

nemakam (changes requested):
Sorry for being blunt, but I don't think this PR is clean in terms of how we are approaching the fix. 
The sessionPump is built in a way that if there is no message received within the `sessionHandlerOptions.MessageWaitTimeout` time, we will close the session and then move on to the next session.
With you current change, it will end up closing and reconnecting the session when there are no messages.

The second problem is how does setting `sessionHandlerOptions.SessionId` interact with `maxConcurrentSession`? 

There was possible solution mentioned here - https://github.com/Azure/azure-service-bus/issues/131#issuecomment-412003895

alberthoekstra:
Okay, sorry for the delay. But I've got it working.

In the MessageSession class, the OnMessageHandler method was overridden from the base class. It just threw an exception saying it's not supported.

I've removed that override and it now simply works like expected.

I'm sure I'm missing something because this is way too simple.
I couldn't find the reason why it threw the exception in the override, maybe one of you can remember?

nemakam:
After fixing the above mentioend issue, @alberthoekstra , could you do a manual test once in your machine and let this Handler run for say 15 minutes when there are messages to be consumed, and another 15 minutes when there are no messages to be consumed. Check for any errors that pop up during that time.

SeanFeldman:
@alberthoekstra it looks like you're very close to complete this PR. Are you sure you don't want to run that manual test to ensure all works?

alberthoekstra:
The manual tests are not the only things @nemakam suggested right?
I think the change i've made should be reverted and there should be an own receive pump somewhere.

SeanFeldman:

axisc:
@alberthoekstra  we're closing this PR given that this repo has been migrated to [Azure SDK for .NET](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/servicebus/Microsoft.Azure.ServiceBus). 

If you feel that you still want to pursue this feature, please open an issue in the new repo and we can discuss this further.

PR #629, opened 2019-01-07
Fix equality check for QueueDescription

mnordlindh:
## Description

The equality check between a manually created created instance of the QueueDescription class and an instance parsed from the XML response by the ManagementClient will not be considered equal. This is an issue when doing operations like create a queue if it is non existing.

The issue is originated in the property AuthorizationRules that was setting the underlying field if it was null. When an instance is being parsed from an non-existing set of rules in the XML response this property is set to an empty AuthorizationRules instance.

I decided to remove the underlying variable for simplicity and set a default value (empty AuthorizationRules instace) to remove the state-changing-getter that just seems to complicate things.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
+++ b/src/Microsoft.Azure.ServiceBus/Management/QueueDescription.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Azure.ServiceBus.Management
         int maxDeliveryCount = 10;
         string forwardTo = null;
         string forwardDeadLetteredMessagesTo = null;
-        AuthorizationRules authorizationRules = null;
         string userMetadata = null;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


nemakam:
The code was written this way to make sure we do not pass authorizationRules xml element when do 
```csharp
var qd = new QueueDescription("myq");
await managementClient.CreateQueue(qd);
```
If we serialize this, since `AuthorizationRules` is not null, we will end up passing `<AuthorizationRules>` element in the description. 
But I would think its okay to pass the empty authorization rules.

mnordlindh:
> The code was written this way to make sure we do not pass authorizationRules xml element when do
> 
> ```cs
> var qd = new QueueDescription("myq");
> await managementClient.CreateQueue(qd);
> ```
> 
> If we serialize this, since `AuthorizationRules` is not null, we will end up passing `<AuthorizationRules>` element in the description.
> But I would think its okay to pass the empty authorization rules.

This was the previous behavior as well because the Serialize method were using the public AuthorizationRules getter that would return a new object if the internal field was null generating a <AuthorizationRules /> XML node in the request. The change will not affect this. Though now you could enforce the removal of the XML node by explicitly setting the AuthorizationRules to null which was not possible before.

> Could you please do the same changes for TopicDescription as well?

Will do!

Issue #582, opened 2018-10-04
Use batching in MessageReceiver.RegisterMessageHandler


WhiteRabbit68:
This is a feature request. Message processing is quite slow using MessageReceiver.RegisterMessageHandler because ReceiveMessagePump deals with one message at a time causing lots of roundtrips completing or abandoning messages, no matter the PreFetchCount.

Concept (BatchingReceiveMessagePump):
``` c#
var batchSize = Math.Min(this.BatchSize, this.receiver.PrefetchCount); 
while (!token.IsCancellationRequested) {
   IList<Message> messages;
   while (!token.IsCancellationRequested && (messages = await this.receiver.ReceiveAsync(batchSize,  this.ReceiveTimeout).ConfigureAwait(false)) != null) {
      if (messages.Count == 0) break;
      var completed = new HashSet<string>();
      var abandoned = new HashSet<string>(messages.Select(m => m.SystemProperties.LockToken));
      foreach (var message in messages) {
         var lockToken = message.SystemProperties.LockToken;
         try {
            await this.handler(message, token).ConfigureAwait(false);
            completed.Add(lockToken);
            abandoned.Remove(lockToken);
         }
         catch (Exception ex) {
            // Exception handler might send to retry queue, deadletter or reschedule
            var handled = await this.exceptionHandler(m, ex).ConfigureAwait(false);
            if (handled) {
               completed.Add(lockToken);
               abandoned.Remove(lockToken);
            }
         } 
      }
      // No batch abandon available, causing slowness when processing fails
      foreach (var lockToken in abandoned)
         await this.receiver.AbandonAsync(lockToken).ConfigureAwait(false);
      if (completed.Any()) // Cannot complete empty list!
         await this.receiver.CompleteAsync(completed).ConfigureAwait(false);
   }
}
```

I made something like this and went from processing around 10 messages/second to 500 messages/second with PreFetchCount around 1000 and BatchSize 500, depending on processing time per message, message size and the lock timeout (allowing larger batches), maybe auto-renewing locks could push this even further but I haven't investigated.

abombss:
This is a big breaking change for us as well when we look at switching clients.  In order to achieve the throughput we are looking for we process batches of messages at a time.  It would be great to extend the interface to accept a Func<Message[]> message array or collection.  I am open to trying a pull request if people think this is a good idea and worth including.

SeanFeldman:
> This is a big breaking change

[...]

PR #486, opened 2018-06-14
Allow clients to report if they own or share the underlying connection string (Part 2)

SeanFeldman:
Follow up for PR #485 which fixed issue #482 

- Extract `OwnsConnection` into `IClientEntity` contract

PR #636, opened 2019-01-23
[WIP] Add support for batched receive via MessageReceiver.RegisterMessageBatchHandler

Einarsson:
add registerMessageHandler on receiver clients that supports batching

This is a proposed change to add batch support to the receiver clients. RegisterMessageHandler currently only processes one message at a time. Batching it would make it easier to handle chunks of messages if you need it to optimize performance towards data sources like SQL etc.

https://github.com/Azure/azure-service-bus-dotnet/issues/582

Changes aren't 100% complete. I have only run manual tests and thus haven't written any unit tests yet. I would like to know if it is at all a viable implementation before proceeding. There might also be formatting issues and other details that might be improved and then I'll happily take the feedback and fix it when I get around to it.

TL;DR; If you like this approach, I can continue working on it with your feedback; else throw it. :)

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/BatchMessageReceivePump.cs
+++ b/src/Microsoft.Azure.ServiceBus/BatchMessageReceivePump.cs
@@ -1,0 +1,277 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.ServiceBus
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Core;
+    using Primitives;
+
+    sealed class BatchMessageReceivePump
+    {
+        readonly Func<IEnumerable<Message>, CancellationToken, Task> onBatchMessageCallback;
+        readonly string endpoint;
+        readonly MessageHandlerOptions registerHandlerOptions;
+        readonly IMessageReceiver messageReceiver;
+        readonly CancellationToken pumpCancellationToken;
+        readonly SemaphoreSlim maxConcurrentCallsSemaphoreSlim;
+        readonly ServiceBusDiagnosticSource diagnosticSource;
+
+        public BatchMessageReceivePump(IMessageReceiver messageReceiver,
+            MessageHandlerOptions registerHandlerOptions,
+            Func<IEnumerable<Message>, CancellationToken, Task> callback,
+            Uri endpoint,
+            CancellationToken pumpCancellationToken)
+        {
+            this.messageReceiver = messageReceiver ?? throw new ArgumentNullException(nameof(messageReceiver));
+            this.registerHandlerOptions = registerHandlerOptions;
+            this.onBatchMessageCallback = callback;
+            this.endpoint = endpoint.Authority;
+            this.pumpCancellationToken = pumpCancellationToken;
+            this.maxConcurrentCallsSemaphoreSlim = new SemaphoreSlim(this.registerHandlerOptions.MaxConcurrentCalls);
+            this.diagnosticSource = new ServiceBusDiagnosticSource(messageReceiver.Path, endpoint);
+        }
+
+        public void StartPump()
+        {
+            TaskExtensionHelper.Schedule(() => this.BatchMessagePumpTaskAsync());
+        }
+
+        bool ShouldRenewLock()
+        {
+            return
+                this.messageReceiver.ReceiveMode == ReceiveMode.PeekLock &&
+                this.registerHandlerOptions.AutoRenewLock;
+        }
+
+        Task RaiseExceptionReceived(Exception e, string action)
+        {
+            var eventArgs = new ExceptionReceivedEventArgs(e, action, this.endpoint, this.messageReceiver.Path, this.messageReceiver.ClientId);
+            return this.registerHandlerOptions.RaiseExceptionReceived(eventArgs);
+        }
+
+        async Task BatchMessagePumpTaskAsync()
+        {
+            while (!this.pumpCancellationToken.IsCancellationRequested)
+            {
+                IEnumerable<Message> messages = null;
+                try
+                {
+                    await this.maxConcurrentCallsSemaphoreSlim.WaitAsync(this.pumpCancellationToken).ConfigureAwait(false);
+                    messages = await this.messageReceiver.ReceiveAsync(this.registerHandlerOptions.MaxMessageCount, this.registerHandlerOptions.ReceiveTimeOut).ConfigureAwait(false);
+
+                    if (messages != null)
+                    {
+                        TaskExtensionHelper.Schedule(() =>
+                        {
+                            if (ServiceBusDiagnosticSource.IsEnabled())
+                            {
+                                return this.BatchMessageDispatchTaskInstrumented(messages);
+                            }
+                            else
+                            {
+                                return this.BatchMessageDispatchTask(messages);
+                            }
+                        });
+                    }
+                }
+                catch (Exception exception)
+                {
+                    // Not reporting an ObjectDisposedException as we're stopping the pump
+                    if (!(exception is ObjectDisposedException && this.pumpCancellationToken.IsCancellationRequested))
+                    {
+                        MessagingEventSource.Log.MessageReceivePumpTaskException(this.messageReceiver.ClientId, string.Empty, exception);
+                        await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.Receive).ConfigureAwait(false);
+                    }
+                }
+                finally
+                {
+                    // Either an exception or for some reason message was null, release semaphore and retry.
+                    if (messages == null)
+                    {
+                        this.maxConcurrentCallsSemaphoreSlim.Release();
+                        MessagingEventSource.Log.MessageReceiverPumpTaskStop(this.messageReceiver.ClientId, this.maxConcurrentCallsSemaphoreSlim.CurrentCount);
+                    }
+                }
+            }
+        }
+
+        async Task BatchMessageDispatchTaskInstrumented(IEnumerable<Message> messages)
+        {
+            IEnumerable<Activity> activities = this.diagnosticSource.ProcessStart(messages);
+            Task processTask = null;
+            try
+            {
+                processTask = BatchMessageDispatchTask(messages);
+                await processTask.ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                this.diagnosticSource.ReportException(e);
+                throw;
+            }
+            finally
+            {
+                this.diagnosticSource.ProcessStop(activities, processTask?.Status);
+            }
+        }
+
+        async Task BatchMessageDispatchTask(IEnumerable<Message> messages)
+        {
+            CancellationTokenSource renewLockCancellationTokenSource = null;
+            Timer autoRenewLockCancellationTimer = null;
+
+            MessagingEventSource.Log.BatchMessageReceiverPumpDispatchTaskStart(this.messageReceiver.ClientId, messages);
+
+            if (this.ShouldRenewLock())
+            {
+                renewLockCancellationTokenSource = new CancellationTokenSource();
+                TaskExtensionHelper.Schedule(() => this.RenewMessageLockTask(messages, renewLockCancellationTokenSource.Token));
+
+                // After a threshold time of renewal('AutoRenewTimeout'), create timer to cancel anymore renewals.
+                autoRenewLockCancellationTimer = new Timer(this.CancelAutoRenewLock, renewLockCancellationTokenSource, this.registerHandlerOptions.MaxAutoRenewDuration, TimeSpan.FromMilliseconds(-1));
+            }
+
+            try
+            {
+                MessagingEventSource.Log.BatchMessageReceiverPumpUserCallbackStart(this.messageReceiver.ClientId);
+                await this.onBatchMessageCallback(messages, this.pumpCancellationToken).ConfigureAwait(false);
+
+                MessagingEventSource.Log.BatchMessageReceiverPumpUserCallbackStop(this.messageReceiver.ClientId);
+            }
+            catch (Exception exception)
+            {
+                MessagingEventSource.Log.BatchMessageReceiverPumpUserCallbackException(this.messageReceiver.ClientId, exception);
+                await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.UserCallback).ConfigureAwait(false);
+
+                // Nothing much to do if UserCallback throws, Abandon message and Release semaphore.
+                if (!(exception is MessageLockLostException))
+                {
+                    await this.AbandonMessagesIfNeededAsync(messages).ConfigureAwait(false);
+                }
+
+                if (ServiceBusDiagnosticSource.IsEnabled())
+                {
+                    this.diagnosticSource.ReportException(exception);
+                }
+                // AbandonMessageIfNeededAsync should take care of not throwing exception
+                this.maxConcurrentCallsSemaphoreSlim.Release();
+
+                return;
+            }
+            finally
+            {
+                renewLockCancellationTokenSource?.Cancel();
+                renewLockCancellationTokenSource?.Dispose();
+                autoRenewLockCancellationTimer?.Dispose();
+            }
+
+            // If we've made it this far, user callback completed fine. Complete message and Release semaphore.
+            await this.CompleteMessagesIfNeededAsync(messages).ConfigureAwait(false);
+            this.maxConcurrentCallsSemaphoreSlim.Release();
+
+            MessagingEventSource.Log.BatchMessageReceiverPumpDispatchTaskStop(this.messageReceiver.ClientId, this.maxConcurrentCallsSemaphoreSlim.CurrentCount);
+        }
+
+        void CancelAutoRenewLock(object state)
+        {
+            var renewLockCancellationTokenSource = (CancellationTokenSource)state;
+            try
+            {
+                renewLockCancellationTokenSource.Cancel();
+            }
+            catch (ObjectDisposedException)
+            {
+                // Ignore this race.
+            }
+        }
+
+        async Task AbandonMessagesIfNeededAsync(IEnumerable<Message> messages)
+        {
+            try
+            {
+                if (this.messageReceiver.ReceiveMode == ReceiveMode.PeekLock)
+                {
+                    var lockTokens = messages.Select(x => x.SystemProperties.LockToken);
+                    await this.messageReceiver.AbandonAsync(lockTokens).ConfigureAwait(false);
+                }
+            }
+            catch (Exception exception)
+            {
+                await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.Abandon).ConfigureAwait(false);
+            }
+        }
+
+        async Task CompleteMessagesIfNeededAsync(IEnumerable<Message> messages)
+        {
+            try
+            {
+                if (this.messageReceiver.ReceiveMode == ReceiveMode.PeekLock &&
+                    this.registerHandlerOptions.AutoComplete)
+                {
+                    var lockTokens = messages.Select(x => x.SystemProperties.LockToken);
+                    await this.messageReceiver.CompleteAsync(lockTokens).ConfigureAwait(false);
+                }
+            }
+            catch (Exception exception)
+            {
+                await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.Complete).ConfigureAwait(false);
+            }
+        }
+
+        async Task RenewMessageLockTask(IEnumerable<Message> messages, CancellationToken renewLockCancellationToken)
+        {
+            while (!this.pumpCancellationToken.IsCancellationRequested &&
+                   !renewLockCancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    var amount = MessagingUtilities.CalculateRenewAfterDuration(messages.Last().SystemProperties.LockedUntilUtc);
+                    MessagingEventSource.Log.BatchMessageReceiverPumpRenewMessageStart(this.messageReceiver.ClientId, amount);
+
+                    // We're awaiting the task created by 'ContinueWith' to avoid awaiting the Delay task which may be canceled
+                    // by the renewLockCancellationToken. This way we prevent a TaskCanceledException.
+                    var delayTask = await Task.Delay(amount, renewLockCancellationToken)
+                        .ContinueWith(t => t, TaskContinuationOptions.ExecuteSynchronously)
+                        .ConfigureAwait(false);
+                    if (delayTask.IsCanceled)
+                    {
+                        break;
+                    }
+
+                    if (!this.pumpCancellationToken.IsCancellationRequested &&
+                        !renewLockCancellationToken.IsCancellationRequested)
+                    {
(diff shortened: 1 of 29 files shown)


src/Microsoft.Azure.ServiceBus/Core/IMessageReceiver.cs
```diff
@@ -133,6 +133,17 @@ public interface IMessageReceiver : IReceiverClient
...
+        /// <summary>
+        /// Renews the locks on the messages. The locks will be renewed based on the setting specified on the queue.
+        /// </summary>
+        /// <remarks>
+        /// When messages are received in <see cref="ServiceBus.ReceiveMode.PeekLock"/> mode, the messages is locked on the server for this
+        /// receiver instance for a duration as specified during the Queue/Subscription creation (LockDuration).
+        /// If processing of the messages requires longer than this duration, the lock needs to be renewed.
+        /// For each renewal, it resets the time the message is locked by the LockDuration set on the Entity.
+        /// </remarks>
+        Task RenewLocksAsync(IEnumerable<Message> messages);
```
SeanFeldman: Note: this will be a breaking change that will require the feature to be released with the next major if feature is accepted.

SeanFeldman: Alternatively, it could be added to `IMessageReceeiver` implementations and `IMessageReceiver` change would be added as a separate PR for the next major, similar to #486.

src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
```diff
@@ -63,6 +63,27 @@ public interface IReceiverClient : IClientEntity
...
+        /// <summary>
+        /// Receive messages continuously from the entity. Registers a message handler and begins a new thread to receive messages.
+        /// This handler(<see cref="Func{Message, CancellationToken, Task}"/>) is awaited on every time a new message is received by the receiver.
+        /// </summary>
+        /// <param name="handler">A <see cref="Func{Message, CancellationToken, Task}"/> that processes messages.</param>
+        /// <param name="exceptionReceivedHandler">A <see cref="Func{T1, TResult}"/> that is invoked during exceptions.
+        /// <see cref="ExceptionReceivedEventArgs"/> contains contextual information regarding the exception.</param>
+        /// <remarks>Enable prefetch to speed up the receive rate.
+        /// Use <see cref="RegisterMessageHandler(Func{Message,CancellationToken,Task}, MessageHandlerOptions)"/> to configure the settings of the pump.</remarks>
+        void RegisterMessageHandler(Func<IEnumerable<Message>, CancellationToken, Task> handler, Func<ExceptionReceivedEventArgs, Task> exceptionReceivedHandler);
```
SeanFeldman: Same comment as for `IMessageReceiver`.

src/Microsoft.Azure.ServiceBus/Core/IReceiverClient.cs
```diff
@@ -63,6 +63,27 @@ public interface IReceiverClient : IClientEntity
...
+
+        /// <summary>
+        /// Receive messages continuously from the entity. Registers a message handler and begins a new thread to receive messages.
+        /// This handler(<see cref="Func{Message, CancellationToken, Task}"/>) is awaited on every time a new message is received by the receiver.
+        /// </summary>
+        /// <param name="handler">A <see cref="Func{Message, CancellationToken, Task}"/> that processes messages.</param>
+        /// <param name="messageHandlerOptions">The <see cref="MessageHandlerOptions"/> options used to configure the settings of the pump.</param>
+        /// <remarks>Enable prefetch to speed up the receive rate.</remarks>
+        void RegisterMessageHandler(Func<IEnumerable<Message>, CancellationToken, Task> handler, MessageHandlerOptions messageHandlerOptions);
```
SeanFeldman: Given it's a batch, I wonder if rather than using `MessageHandlerOptions` it should be something that would indicate it's a batch. @nemakam thoughts?

src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs
```diff
@@ -75,6 +76,23 @@ public TimeSpan MaxAutoRenewDuration
             }
         }
 
+        /// <summary>Gets or sets the maximum number of messages the message pump should receive.</summary>
+        /// <value>The maximum number of messages.</value>
+        public int MaxMessageCount
```
SeanFeldman: For a single message handler `MaxMessageCount` will be meaningless.
I'm concerned about the confusion it might introduce if the same `MessageHandlerOptions` is re-used for a single message handler and a batch handler.

src/Microsoft.Azure.ServiceBus/BatchMessageReceivePump.cs
```diff
@@ -0,0 +1,277 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Core;
+    using Primitives;
+
+    sealed class BatchMessageReceivePump
```
SeanFeldman: How about `MessageBatchReceiverPump` to be consistent with a parallel PR #539 ?

src/Microsoft.Azure.ServiceBus/Resources.resx
```diff
@@ -216,4 +216,7 @@
   <data name="TimeoutMustBePositiveNonZero" xml:space="preserve">
     <value>Argument {0} must be a positive non-zero timeout value. The provided value was {1}.</value>
   </data>
+  <data name="MaxMessageCountMustBeGreaterThanZero1" xml:space="preserve">
+    <value>The specified value '{0}' is invalid. "maxMessageCount" must be greater than zero.</value>
```
SeanFeldman:
```suggestion
    <value>The specified value '{0}' is invalid. "MaxMessageCount" must be greater than zero.</value>
```
Also, I think the type including `MaxMessageCount` should be specified to ensure it's a clear exception message and users should know what to change.

src/Microsoft.Azure.ServiceBus/ServiceBusDiagnosticsSource.cs
```diff
@@ -17,7 +17,7 @@ internal class ServiceBusDiagnosticSource
         public const string BaseActivityName = "Microsoft.Azure.ServiceBus.";
 
         public const string ExceptionEventName = BaseActivityName + "Exception";
-        public const string ProcessActivityName =  BaseActivityName + "Process";
+        public const string ProcessActivityName = BaseActivityName + "Process";
```
SeanFeldman: Looks like this was the only change that triggered some whitespaces cleanup 🙂

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/QueueClientDiagnosticsTests.cs
```diff
@@ -28,7 +28,7 @@ async Task EventsAreNotFiredWhenDiagnosticsIsDisabled()
             this.listener.Disable();
 
             await TestUtility.SendMessagesAsync(this.queueClient.InnerSender, 1);
-            this.queueClient.RegisterMessageHandler((msg, ct) => {
+            this.queueClient.RegisterMessageHandler((Message msg, CancellationToken ct) => {
```
SeanFeldman: I'm not sure we need to spell out the types. Though if you do want to provide better names for the arguments to indicate what they stand for then 👍

Also, this kind of change would be probably better to have as a separate PR to separate the concerns and leave the focus on batched receive.

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/QueueClientDiagnosticsTests.cs
```diff
@@ -54,7 +54,7 @@ async Task EventsAreNotFiredWhenDiagnosticsIsDisabledForQueue()
                 queueName == null || queueName.ToString() != TestConstants.NonPartitionedQueueName);
 
             await TestUtility.SendMessagesAsync(this.queueClient.InnerSender, 1);
-            this.queueClient.RegisterMessageHandler((msg, ct) =>
+            this.queueClient.RegisterMessageHandler((Message msg, CancellationToken ct) =>
```
SeanFeldman: Same as above

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/QueueClientDiagnosticsTests.cs
```diff
@@ -86,7 +86,7 @@ async Task SendAndHandlerFireEvents()
             await TestUtility.SendSessionMessagesAsync(this.queueClient.InnerSender, 1, 1);
             parentActivity.Stop();
 
-            this.queueClient.RegisterMessageHandler((msg, ct) =>
+            this.queueClient.RegisterMessageHandler((Message msg, CancellationToken ct) =>
```
SeanFeldman: Same as above

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/QueueClientDiagnosticsTests.cs
```diff
@@ -151,7 +151,7 @@ async Task SendAndHandlerFireExceptionEvents()
             this.listener.Enable((name, queueName, arg) => !name.EndsWith(".Start") && !name.Contains("Receive") );
 
             int count = 0;
-            this.queueClient.RegisterMessageHandler((msg, ct) =>
+            this.queueClient.RegisterMessageHandler((Message msg, CancellationToken ct) =>
```
SeanFeldman: Same as above

test/Microsoft.Azure.ServiceBus.UnitTests/Diagnostics/QueueClientDiagnosticsTests.cs
```diff
@@ -451,7 +451,7 @@ async Task SendAndHandlerFilterOutStartEvents()
             this.listener.Enable((name, queueName, arg) => !name.EndsWith("Start") && !name.Contains("Receive") && !name.Contains("Exception"));
 
             await TestUtility.SendMessagesAsync(this.queueClient.InnerSender, 1);
-            this.queueClient.RegisterMessageHandler((msg, ct) =>
+            this.queueClient.RegisterMessageHandler((Message msg, CancellationToken ct) =>
```
SeanFeldman: Same as above

test/Microsoft.Azure.ServiceBus.UnitTests/OnMessageQueueTests.cs
```diff
@@ -68,7 +69,7 @@ async Task OnMessageExceptionHandlerCalledTest()
 
             var queueClient = new QueueClient(TestUtility.NamespaceConnectionString, queueName, ReceiveMode.ReceiveAndDelete);
             queueClient.RegisterMessageHandler(
-                (message, token) => throw new Exception("Unexpected exception: Did not expect messages here"),
+                (Message message, CancellationToken token) => throw new Exception("Unexpected exception: Did not expect messages here"),
```
SeanFeldman: Same as above

SeanFeldman (dismissed):
I like the idea. There are a few things to consider. Some are in the comments.
- Focus on Batched receive pump and not cleanups
- Separation of options for a single message vs message batch
- Minor vs major release

Also, @nemakam please have a look.

Einarsson:
Thanks for inputs @SeanFeldman!

One thing though; I havn't figured out a way to nest generics in XML comments; Take this one for example:

`/// Use <see cref="RegisterMessageHandler(Func{Message,CancellationToken,Task}, MessageHandlerOptions)"/> to configure the settings of the pump.</remarks>`

It says incorrect syntax if you do RegisterMessageHandler(Func{List{Message},CancellationToken,Task}


I'll look into the changes you've proposed. :) Should also be possible to avoid an unnecessary breaking change.

Einarsson:
Interesting.

This works:
`/// Use <see cref="RegisterBatchMessageHandler(Func{IList{Message},CancellationToken,Task}, BatchMessageHandlerOptions)"/> to configure the settings of the pump.</remarks>`

This does not:
`/// This handler(<see cref="Func{IList{Message}, CancellationToken, Task}"/>>) is awaited on every time a new message is received by the receiver.`

It gives: 
CS1584 C# XML comment has syntactically incorrect cref attribute
CS1658 C# Type parameter declaration must be an identifier not a type. See also error CS0081.

Appearently a known problem:
https://stackoverflow.com/questions/684982/referring-to-a-generic-type-of-a-generic-type-in-c-sharp-xml-documentation/797364

Einarsson:
Sorted out lot of proposed changes.

- Renamed to MessageBatch to get naming consistent with #539
- Remove additions to core interfaces, will add in separate PR later on
- Separated HandlerOptions classes
- Revert VS accidental formatting cleanups :)

However issue with the XML comments still persists and are unsolvable, not sure how to proceed with that. It could work if you introduce some MessageBatch object instead of IList<Message> but the idea of XML comments dictating code structure sickens me. Might be viable if you'd want some kind of metadata regarding the messages included in the batch, otherwise its just unnecessary clutter.

SeanFeldman:
The build is OK with XMLDocs, but it's failing on tests. Could you please verify those?

Einarsson:
Ran the testa successfully yesterday. I can double check later. 

The build is ok with XML comments because i havnt changed to List<Message> yet. Seems to be no real way around it.

Einarsson:
Tests runnin. :)

Einarsson:
Will do @SeanFeldman ! Gonna research the existing tests to setup a test in a similar fashion. :)

SeanFeldman:
You will need an environment variable with connection string to the namespace with a set of predefined entities. All this is documented [here](https://github.com/Azure/azure-service-bus-dotnet#how-do-i-run-the-unit-tests).

Einarsson:
I've added some tests that uses the subscription client and messageBatchHandler. I've tested it manually with console apps as well before where I sent 10 000 or more messages in batches of max 100.

test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
```diff
@@ -511,6 +526,7 @@ namespace Microsoft.Azure.ServiceBus.Core
         int PrefetchCount { get; set; }
         Microsoft.Azure.ServiceBus.ReceiveMode ReceiveMode { get; }
         System.Threading.Tasks.Task AbandonAsync(string lockToken, System.Collections.Generic.IDictionary<string, object> propertiesToModify = null);
+        System.Threading.Tasks.Task AbandonAsync(System.Collections.Generic.IList<string> lockTokens, System.Collections.Generic.IDictionary<string, object> propertiesToModify = null);
```
SeanFeldman: Attention: breaking change!

If you want to add this to the current major, the change to the contract should be raised as a follow up PR.
Similar to what PR https://github.com/Azure/azure-service-bus-dotnet/pull/486 does.

If the change is added in this PR, this will force a major increase and could delay merge until major version is considered.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1263,6 +1405,81 @@ protected virtual async Task<DateTime> OnRenewLockAsync(string lockToken)
...
+                    amqpRequestMessage.AmqpMessage.ApplicationProperties.Map[ManagementConstants.Request.AssociatedLinkName] = receiveLink.Name;
+                }
+                amqpRequestMessage.Map[ManagementConstants.Properties.LockTokens] = new[] { lockTokens.Select(lockToken => new Guid(lockToken)) };
+
+                var amqpResponseMessage = await this.ExecuteRequestResponseAsync(amqpRequestMessage).ConfigureAwait(false);
+
+                if (amqpResponseMessage.StatusCode == AmqpResponseStatusCode.OK)
+                {
+                    var lockedUntilUtcTimes = amqpResponseMessage.GetValue<IEnumerable<DateTime>>(ManagementConstants.Properties.Expirations);
+                    lockedUntilUtc = lockedUntilUtcTimes.Last();
```
SeanFeldman: Given there are multiple messages in the batch, there could should it be the last or the first renewed message's date/time?
Or perhaps the earliest expiring date/time?

SeanFeldman (changes requested):

src/Microsoft.Azure.ServiceBus/MessageBatchReceivePump.cs
```diff
@@ -0,0 +1,277 @@
...
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Core;
+    using Primitives;
+
+    sealed class MessageBatchReceivePump
+    {
```
nemakam: There's a LOT of duplication here. Since this is an internal class, we don't need it to be explicit about batching. I'd recommend doing the changes directly in the base class by handling all the cases.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1198,6 +1330,16 @@ protected virtual Task OnAbandonAsync(string lockToken, IDictionary<string, obje
             return this.DisposeMessagesAsync(lockTokens, GetAbandonOutcome(propertiesToModify));
         }
 
+        protected virtual Task OnAbandonAsync(IEnumerable<string> lockTokens, IDictionary<string, object> propertiesToModify = null)
```
nemakam: >OnAbandonAsync [](start = 31, length = 14)

You should test these.

src/Microsoft.Azure.ServiceBus/Core/MessageReceiver.cs
```diff
@@ -1263,6 +1405,81 @@ protected virtual async Task<DateTime> OnRenewLockAsync(string lockToken)
             return lockedUntilUtc;
         }
 
+        protected virtual async Task<DateTime> OnRenewLocksAsync(IEnumerable<string> lockTokens)
```
nemakam: >OnRenewLocksAsync [](start = 47, length = 17)

You should test these. I am not sure if this works from service side.

nemakam: What about cases when in a batch of 100 messages, 50 are already consumed (or completed) and you are renewing all 100. ?

---
In reply to: [255765849](https://github.com/Azure/azure-service-bus-dotnet/pull/636#discussion_r255765849) [](ancestors = 255765849)

nemakam: Do not proceed till you've tested these base foundations as I feel you might get blocked somehwere. 

---
In reply to: [255766004](https://github.com/Azure/azure-service-bus-dotnet/pull/636#discussion_r255766004) [](ancestors = 255766004,255765849)

src/Microsoft.Azure.ServiceBus/MessageBatchHandlerOptions.cs
```diff
@@ -0,0 +1,115 @@
...
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Primitives;
+    using System.Collections.Generic;
+
+    /// <summary>Provides options associated with message pump processing using
+    /// <see cref="QueueClient.RegisterMessageBatchHandler(Func{IList{Message}, CancellationToken, Task}, MessageBatchHandlerOptions)" /> and
+    /// <see cref="SubscriptionClient.RegisterMessageBatchHandler(Func{IList{Message}, CancellationToken, Task}, MessageBatchHandlerOptions)" />.</summary>
+    public sealed class MessageBatchHandlerOptions
```
nemakam: >MessageBatchHandlerOptions [](start = 24, length = 26)

This should just derive from the non-batch handler and add the extra property.

src/Microsoft.Azure.ServiceBus/MessageBatchReceivePump.cs
```diff
@@ -0,0 +1,277 @@
...
+        async Task MessageBatchPumpTaskAsync()
+        {
+            while (!this.pumpCancellationToken.IsCancellationRequested)
+            {
+                IList<Message> messages = null;
+                try
+                {
+                    await this.maxConcurrentCallsSemaphoreSlim.WaitAsync(this.pumpCancellationToken).ConfigureAwait(false);
+                    messages = await this.messageReceiver.ReceiveAsync(this.registerHandlerOptions.MaxMessageCount, this.registerHandlerOptions.ReceiveTimeOut).ConfigureAwait(false);
+
```
nemakam: Sorry to break this to you, but receiveing in a batch still doesn't work as expected and its a very high possibliity that you will end up with a batch size of 1 always. It needs some changes across client and service

src/Microsoft.Azure.ServiceBus/MessageBatchReceivePump.cs
```diff
@@ -0,0 +1,277 @@
...
+            catch (Exception exception)
+            {
+                MessagingEventSource.Log.MessageBatchReceiverPumpUserCallbackException(this.messageReceiver.ClientId, exception);
+                await this.RaiseExceptionReceived(exception, ExceptionReceivedEventArgsAction.UserCallback).ConfigureAwait(false);
+
+                // Nothing much to do if UserCallback throws, Abandon message and Release semaphore.
+                if (!(exception is MessageLockLostException))
+                {
+                    await this.AbandonMessagesIfNeededAsync(messages).ConfigureAwait(false);
+                }
```
nemakam: You are abondoning every message because one message ran into lock lost exception?

src/Microsoft.Azure.ServiceBus/MessageBatchReceivePump.cs
```diff
@@ -0,0 +1,277 @@
...
+
+            try
+            {
+                MessagingEventSource.Log.MessageBatchReceiverPumpUserCallbackStart(this.messageReceiver.ClientId);
+                await this.onMessageBatchCallback(messages, this.pumpCancellationToken).ConfigureAwait(false);
+
+                MessagingEventSource.Log.MessageBatchReceiverPumpUserCallbackStop(this.messageReceiver.ClientId);
+            }
+            catch (Exception exception)
+            {
```
nemakam: How do you know which message ran into the exception?

nemakam (changes requested):
I have left couple of comments to think about. I'm sorry that I'm adding comments after 13 commits, but I dont think this is going to work out easily. Adding such batch scenarios is very tricky to get it right. 
I do not have a lot of suggestions as I'm not sure what the best API would be that would not break too many things. However, the current API that you are exposing might not work out.

src/Microsoft.Azure.ServiceBus/MessageBatchHandlerOptions.cs
SeanFeldman: Not a big fan of inheritance. A slight change will cause waves throughout all derived classes.

SeanFeldman:

src/Microsoft.Azure.ServiceBus/MessageBatchReceivePump.cs
Einarsson: I havn't done that much testing but I've tried sending 10 000 messages, max message count of 500 with prefetch 1000. I received 300-500 messages each pump. Can look into it more deeply though when I get the time.

axisc:
@Einarsson we're closing this PR given that this repo has been migrated to [Azure SDK for .NET](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/servicebus/Microsoft.Azure.ServiceBus). 

If you feel that you still want to pursue this feature, please open an issue in the new repo and we can discuss this further.

PR #638, opened 2019-01-24
Fix NamespaceClient.GetNamespaceInfoAsync()

### LINKED ISSUE #637: Management client doesn't return correct NamespaceInfo
Two issues:

1. `ManagementClient.GetNamespaceInfoAsync()` invocation for premium tier never returns number of MUs.
2. `NamespaceType` is always `NamespaceType.Others` as broker returns `Messaging` for ASB namespaces, not `ServiceBus`.

This is a broken functionality as it doesn't do what it supposed.

Tested version: 3.3.0

Running simple test on standard and premium namespaces returns the following:

image

SeanFeldman:
Fixes #637

### DIFF
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
@@ -730,6 +730,7 @@ namespace Microsoft.Azure.ServiceBus.Management
     public enum NamespaceType
     {
         ServiceBus = 0,
+        Messaging = 1,
         Mixed = 2,
         Others = 99,
     }
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
@@ -730,7 +730,6 @@ namespace Microsoft.Azure.ServiceBus.Management
     public enum NamespaceType
     {
         ServiceBus = 0,
-        Messaging = 1,
         Mixed = 2,
         Others = 99,
     }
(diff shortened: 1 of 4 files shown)


SeanFeldman:
This will need to be a **minor** release as we're adding to the public API of an existing type `NamespaceType`.

SeanFeldman:
@nemakam should we deprecate `NamespaceType.ServiceBus` or leave it as-is?

test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt
```diff
@@ -730,6 +730,7 @@ namespace Microsoft.Azure.ServiceBus.Management
     public enum NamespaceType
     {
         ServiceBus = 0,
+        Messaging = 1,
```
nemakam: This is actually not true. Problem is that `ServiceBus` is called `Messaging` in service side. Didnt realize when I renamed the enum. The correct thing is to rename `ServiceBus` to `Messaging`. But that is going to break things. So the alternative that I can think of is to manually handle the deserialization and if it is Messaging, then convert it to ServiceBus. :(

SeanFeldman:
I can live with that, but we'll need to change that in v4.

Sean Feldman

SeanFeldman:
@nemakam updated

SeanFeldman:
@nemakam could you please raise an issue for 4.0.0 milestone to address https://github.com/Azure/azure-service-bus-dotnet/pull/638/commits/f10cae959227ab8d28e9fe5877b968b789a1df0e#diff-4f0427b57425e52bf6ea1d62facac6b9R69? Thanks.

SeanFeldman:
@nemakam please review

SeanFeldman:
@nemakam who could review this?

nemakam:
I guess we could just leave it as-is.

---
In reply to: [457061141](https://github.com/Azure/azure-service-bus-dotnet/pull/638#issuecomment-457061141) [](ancestors = 457061141)

PR #641, opened 2019-02-06
Do not allow sending empty collection

SeanFeldman:
Fixes #550

PR #642, opened 2019-02-07
Allow sending empty collection by skipping the actual send

### LINKED ISSUE #550: ISenderClient:SendAsync(IList<Message> messageList) throws null reference when messageList is empty
## Actual Behavior
1. SendAsync throws null reference exception when given collection of messages is empty.

## Expected Behavior
1. No exception should be thrown.

## Simplest repro
```csharp
        [Fact]
        [DisplayTestMethodName]
        void Empty_Collection_Of_Messages_Returns_Empty_AMQP_Message()
        {
            var messages = new Message[0];

            var batchedMessage = AmqpMessageConverter.BatchSBMessagesAsAmqpMessage(messages);

            Assert.NotNull(batchedMessage);
        }
```

> System.NullReferenceException : Object reference not set to an instance of an object.
>    at Microsoft.Azure.ServiceBus.Amqp.AmqpMessageConverter.BatchSBMessagesAsAmqpMessage(IEnumerable`1 sbMessages)
>    at Microsoft.Azure.ServiceBus.UnitTests.AmqpConverterTests.Empty_Collection_Of_Messages_Returns_Empty_AMQP_Message()

## Reason
https://github.com/Azure/azure-service-bus-dotnet/blob/6f144e91310dcc7bd37aba4e8aebd535d13fa31a/src/Microsoft.Azure.ServiceBus/Amqp/AmqpMessageConverter.cs#L59
`dataList` is initalized in a foreach loop, so when there is no messages it will be null.
Later one, its used by `amqpMessage = AmqpMessage.Create(dataList);` what causes null reference exception.

## Versions  
- OS platform and version: Windows 10
- .NET Version: .NET Core 2.1
- NuGet package version or commit ID: 3.1.0-preview, master

I can pick up this issue if you want.

SeanFeldman:
Fixes #550

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -237,6 +237,11 @@ namespace Microsoft.Azure.ServiceBus.Core
             this.ThrowIfClosed();
 
             var count = MessageSender.ValidateMessages(messageList);
+            if (count <= 0)
+            {
+                return;
+            }
+
             MessagingEventSource.Log.MessageSendStart(this.ClientId, count);
 
             bool isDiagnosticSourceEnabled = ServiceBusDiagnosticSource.IsEnabled();
@@ -451,6 +456,11 @@ namespace Microsoft.Azure.ServiceBus.Core
             if (messageList == null)
             {
                 throw Fx.Exception.ArgumentNull(nameof(messageList));
+            }
+
+            if (messageList.Count == 0)
+            {
+                return count;
             }
 
             foreach (var message in messageList)
--- a/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
+++ b/src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
@@ -458,11 +458,6 @@ namespace Microsoft.Azure.ServiceBus.Core
                 throw Fx.Exception.ArgumentNull(nameof(messageList));
             }
 
-            if (messageList.Count == 0)
-            {
-                return count;
-            }
-
             foreach (var message in messageList)
             {
                 count++;
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
@@ -463,7 +463,7 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
         [Theory]
         [InlineData(TestConstants.NonPartitionedQueueName)]
         [DisplayTestMethodName]
-        async Task MessageSenderThrowsWhenSendingEmptyCollection(string queueName)
+        async Task MessageSenderShouldNotThrowWhenSendingEmptyCollection(string queueName)
         {
             var sender = new MessageSender(TestUtility.NamespaceConnectionString, queueName);
 
--- a/test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
+++ b/test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
@@ -466,8 +466,19 @@ namespace Microsoft.Azure.ServiceBus.UnitTests
         async Task MessageSenderShouldNotThrowWhenSendingEmptyCollection(string queueName)
         {
             var sender = new MessageSender(TestUtility.NamespaceConnectionString, queueName);
+            var receiver = new MessageReceiver(TestUtility.NamespaceConnectionString, queueName, ReceiveMode.ReceiveAndDelete);
 
-            await sender.SendAsync(new List<Message>());
+            try
+            {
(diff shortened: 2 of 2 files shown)


test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
```diff
@@ -459,5 +459,16 @@ public async Task ClientsUseGlobalConnectionCloseFirstClientSecoundClientShouldS
                 await receiver.CloseAsync().ConfigureAwait(false);
             }
         }
+
+        [Theory]
+        [InlineData(TestConstants.NonPartitionedQueueName)]
+        [DisplayTestMethodName]
+        async Task MessageSenderThrowsWhenSendingEmptyCollection(string queueName)
```
thomaslevesque: You might want to rename this method to reflect what is actually being tested

SeanFeldman: WIP. Also, I'm thinking the test is not good enough 🙂

SeanFeldman:
@nemakam please review

test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
nemakam: Yeah the test somehow doesnt look perfect. Why do we even need a receiver here? Of course there is no message sent as its an empty list. Its not worth spending those 3 seconds * number of build configurations to test this.

---
In reply to: [254793100](https://github.com/Azure/azure-service-bus-dotnet/pull/642#discussion_r254793100) [](ancestors = 254793100)

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
```diff
@@ -453,6 +458,11 @@ static int ValidateMessages(IList<Message> messageList)
                 throw Fx.Exception.ArgumentNull(nameof(messageList));
             }
 
+            if (messageList.Count == 0)
```
nemakam: >if (messageList.Count == 0) [](start = 12, length = 27)

This isnt required

test/Microsoft.Azure.ServiceBus.UnitTests/SenderReceiverTests.cs
SeanFeldman: Any ideas how to test it better? I'm open for suggestions.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
SeanFeldman: As in let `foreach` run with zero iterations? Could do that, but then it's less obvious what will happen, no?

SeanFeldman:

SeanFeldman:

dlstucki:
@nemakam, et. al., what's the explanation for not throwing ArgumentException instead? Is it ease of the implementation of this library method? Most APIs I can think of which have some kind of "Send" method tend to throw if you tell it to send but don't give any data for sending. I feel like silently doing nothing might be hiding errors in the code calling this library.

If it's ease of implementation (like avoiding enumerating something twice) could something like this be done?
```
SendBatch(IEnumerable<Whatever> batch)
{
    bool foundAny = false;
    foreach (Whatever w in batch)
    {
        foundAny = true;
        // ... Do something ...
    }

    if (!foundAny) { throw new ArgumentException("batch", "Batch contained no items");
}
```

SeanFeldman:
@dlstucki that was my [initial take](https://github.com/Azure/azure-service-bus-dotnet/pull/641), but there was a compelling argument why an empty list should not throw: https://github.com/Azure/azure-service-bus-dotnet/issues/550#issuecomment-461330073

dlstucki:
> @dlstucki that was my [initial take]

@SeanFeldman Thanks! Seeing other APIs which behave this way makes me feel better.

SeanFeldman:
The only reason this is still not merged is because I have no idea what would be a better way to test this behaviour w/o hitting the broker. If you have ideas @dlstucki, please share 🙂

dlstucki:
@SeanFeldman, I haven't been following along very closely so I probably don't understand it. What do you need to verify in order to test this?  Would this work?
- Sending 1 message before
- Send an empty batch (make sure it doesn't throw)
- Sending 1 message after
- Verify the before and after messages are received successfully

SeanFeldman:
@dlstucki that's sort of [in place](https://github.com/Azure/azure-service-bus-dotnet/pull/642/files#diff-649e3f0fe912cbbd18f4ba770db10005R466) already.

src/Microsoft.Azure.ServiceBus/Core/MessageSender.cs
SeanFeldman: Done

SeanFeldman:
I'll go with this for now and if a better test idea emerges, can be PR-ed

PR #644, opened 2019-02-08
Add support for configuring OperationTimeout in the connection string

0xced:
Like in [Microsoft.ServiceBus.Messaging (.NET Framework)][1], it's now possible to configure the `OperationTimeout` in the connection string.

The `ServiceBusConnection` class constructors are adapted to read the `OperationTimeout` from the connection string and the constructor explicitly having a `TimeSpan operationTimeout` argument is obsoleted with this message:
> Please use the constructor with (string namespaceConnectionString, RetryPolicy retryPolicy) arguments and define the operationTimeout in the connection string.

Also, use the operation timeout defined in the connection string for the ManagementClient and **actually** use the operation timeout as a timeout for the `HttpClient` responsible for the management operations instead of storing the `Constants.DefaultOperationTimeout` in a private field which is never used.

[1]: https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.servicebusconnectionstringbuilder.operationtimeout

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.ServiceBus
         /// <param name="operationTimeout">Duration after which individual operations will timeout.</param>
         /// <param name="retryPolicy">Retry policy for operations. Defaults to <see cref="RetryPolicy.Default"/></param>
         /// <remarks>It is the responsibility of the user to close the connection after use through <see cref="CloseAsync"/></remarks>
-        [Obsolete("Please use the constructor with (string namespaceConnectionString, RetryPolicy retryPolicy) arguments and define the operationTimeout in the connection string.")]
+        [Obsolete("This constructor is obsolete. Use ServiceBusConnection(string namespaceConnectionString, RetryPolicy retryPolicy) constructor instead, providing operationTimeout in the connection string.")]
         public ServiceBusConnection(string namespaceConnectionString, TimeSpan operationTimeout, RetryPolicy retryPolicy = null)
             : this(retryPolicy)
         {
@@ -87,7 +87,7 @@ namespace Microsoft.Azure.ServiceBus
             }
 
             this.InitializeConnection(serviceBusConnectionStringBuilder);
-            // operationTimeout argument wins over OperationTimeout defined in the connection string since it's explicitly specified in the constructor
+            // operationTimeout argument explicitly provided by caller should take precedence over OperationTimeout found in the connection string.
             this.OperationTimeout = operationTimeout;
         }
 
--- a/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Azure.ServiceBus
         /// <param name="namespaceConnectionString">Namespace connection string</param>
         /// <remarks>It is the responsibility of the user to close the connection after use through <see cref="CloseAsync"/></remarks>
         public ServiceBusConnection(string namespaceConnectionString)
-            : this(namespaceConnectionString, Constants.DefaultOperationTimeout, RetryPolicy.Default)
+            : this(namespaceConnectionString, RetryPolicy.Default)
         {
         }
 
@@ -45,11 +45,10 @@ namespace Microsoft.Azure.ServiceBus
         /// Creates a new connection to service bus.
         /// </summary>
         /// <param name="namespaceConnectionString">Namespace connection string.</param>
-        /// <param name="operationTimeout">Duration after which individual operations will timeout.</param>
         /// <param name="retryPolicy">Retry policy for operations. Defaults to <see cref="RetryPolicy.Default"/></param>
         /// <remarks>It is the responsibility of the user to close the connection after use through <see cref="CloseAsync"/></remarks>
-        public ServiceBusConnection(string namespaceConnectionString, TimeSpan operationTimeout, RetryPolicy retryPolicy = null)
-            : this(operationTimeout, retryPolicy)
+        public ServiceBusConnection(string namespaceConnectionString, RetryPolicy retryPolicy = null)
+            : this(retryPolicy)
         {
             if (string.IsNullOrWhiteSpace(namespaceConnectionString))
             {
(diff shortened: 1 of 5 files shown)


0xced:
> Public API has to be updated for changes.

What does this mean? What do I have to do in addition to document the new `ServiceBusConnection` constructor and the new `OperationTimeout` property on the `ServiceBusConnectionStringBuilder` class? I don't see a CHANGELOG.md file, I don't understand what changes you are requesting. 😕

SeanFeldman:
One of the tests, [`APIApprovals`](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/test/Microsoft.Azure.ServiceBus.UnitTests/API/APIApprovals.cs) is failing. That's because the public API has changed and hasn't been approved. To do so, you'll need to run the test and update [`ApiApprovals.ApproveAzureServiceBus.approved.txt`](https://github.com/Azure/azure-service-bus-dotnet/blob/dev/test/Microsoft.Azure.ServiceBus.UnitTests/API/ApiApprovals.ApproveAzureServiceBus.approved.txt) filed with the new content. This will also help to determine if there are any breaking changes or not. So far it looks like a new minor 🙂

0xced:
It wasn't very clear from the [failure][1] since the error was about line ending mismatch:
```
Assert.Equal() Failure
                                 ↓ (pos 443)
Expected: ···edf0bdbccafadfeb6")]\n[assembly: System.Runtime.InteropService···
Actual:   ···edf0bdbccafadfeb6")]\r\n[assembly: System.Runtime.InteropServic···
                                 ↑ (pos 443)
```

Anyway, I have updated the API approvals and the test is now passing.

[1]: https://ci.appveyor.com/project/vinaysurya/azure-service-bus-dotnet/builds/22227559/job/t9o5odojdyjf55kb/tests

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
```diff
@@ -37,19 +37,43 @@ public ServiceBusConnection(ServiceBusConnectionStringBuilder connectionStringBu
...
+        }
+
         /// <summary>
         /// Creates a new connection to service bus.
         /// </summary>
         /// <param name="namespaceConnectionString">Namespace connection string.</param>
         /// <param name="operationTimeout">Duration after which individual operations will timeout.</param>
         /// <param name="retryPolicy">Retry policy for operations. Defaults to <see cref="RetryPolicy.Default"/></param>
         /// <remarks>It is the responsibility of the user to close the connection after use through <see cref="CloseAsync"/></remarks>
+        [Obsolete("Please use the constructor with (string namespaceConnectionString, RetryPolicy retryPolicy) arguments and define the operationTimeout in the connection string.")]
```
SeanFeldman: How about the following:

> This constructor is obsolete. Use ServiceBusConnection(string namespaceConnectionString, RetryPolicy retryPolicy) constructor instead, providing operationTimeout in the connection string.")]

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
```diff
@@ -63,6 +87,8 @@ public ServiceBusConnection(string namespaceConnectionString, TimeSpan operation
             }
 
             this.InitializeConnection(serviceBusConnectionStringBuilder);
+            // operationTimeout argument wins over OperationTimeout defined in the connection string since it's explicitly specified in the constructor
```
SeanFeldman: How about

> operationTimeout argument explicitly provide by caller should take precedence over OperationTimeout found in the connection string.

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
SeanFeldman: How about

> This constructor is obsoleted. Use ServiceBusConnection(string namespaceConnectionString, RetryPolicy retryPolicy)  constructor providing operationTimeout using connection string.

SeanFeldman: How about:

> This constructor is obsoleted. Use ServiceBusConnection(string namespaceConnectionString, RetryPolicy retryPolicy)  constructor providing operationTimeout using connection string.

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
SeanFeldman: How about
> operationTimeout argument explicitly provide by caller should take precedence over OperationTimeout found in the connection string.

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -319,6 +332,17 @@ void ParseConnectionString(string connectionString)
                         this.TransportType = transportType;
                     }
                 }
+                else if (key.Equals(OperationTimeoutConfigName, StringComparison.OrdinalIgnoreCase))
+                {
+                    try
+                    {
+                        this.OperationTimeout = TimeSpan.Parse(value);
```
dlstucki: Does everyone like using a TimeSpan format?
What does this mean?
OperationTimeout=60

What format do SQL connection strings use? (Seconds or something?)

SeanFeldman: > What format do SQL connection strings use? (Seconds or something?)

SQL Server connection timeout is in seconds ([doco](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection.connectiontimeout?view=netframework-4.7.2))

I'd vote to replace `TimeSpan.Parse()` with `TimeSpan.TryParse()` and then guard against values that are not reasonable/out of the acceptable range.

0xced: I think it would be best to use seconds, just like SQL Server indeed. I like the idea to keep it as a `TimeSpan` instead of `int` since it’s self documenting.  I also like the idea to guard against values that are not reasonable. But how do we define this reasonable range? From 1 second to 1 hour? Any other suggestion?

SeanFeldman: @nemakam what would be a reasonable value range in this case, can you suggest?

SeanFeldman:

SeanFeldman:
> BTW, I was recommending only the string representation use number of seconds, the SBConnectionStringBuilder.OperationTimeout should remain a TimeSpan for programmer ease.

@dlstucki yes, that's what everyone agreed upon. 👍

SeanFeldman:
@0xced could you please 
- Sign CLA to unblock merging
- Remove WIP from the title if you're no longer planning to modify anything
- Rebase to ensure it's all good

0xced:
It's still a WIP, I'm waiting on an answer about the *reasonable value range* that should be accepted for the timeout. Also, I have signed the CLA, this is probably a glitch that will disappear the next time I push changes.

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
SeanFeldman: @axisc could you please help to unblock this PR? It's almost done and just requires some feedback from the team. Thank you.

nemakam: Its a fair question. Do we keep the `ToString()` in the `TimeSpan` format, or numberOfSeconds format. Keeping in `TimeSpan` format makes it interchangable with older client, and keeping it numberOfSeconds will make it easier across clients (not just .net and java, but also the other languages like python, node, go etc etc). 
Let me discuss with the broader team and see what they think about this...

nemakam: So here's the current status -
Old .net client uses Timespan.Parse() and Timespan.ToString() , where format is 00:00:60.
Java client uses Duration, which has its own format.
Moving forward, `seconds` feel like a safe bet as the format for operation timeout. So how about this - 

while parsing, we do
```csharp
if (int.TryParse(operationTimeoutInputValue, out var ot))
{
    operationTimeout = TimeSpan.FromSeconds(ot);
}
else
{
    TimeSpan.TryParse(operationTimeoutInputValue, out operationTimeout);
}
```

While doing ToString(), use `operationTimeout.TotalSeconds`.

The outcome is that in new world we will end up using seconds to keep it less confusing about how to parse, but we also let users migrate from old .net client to new .net client.

nemakam:
And for _reasonable value range_ , upper limit of 1 hour seems fine.

0xced:
> lets not have a lower limit

How would the system behave if the timeout is set to zero seconds? Wouldn't it be a way to shoot yourself in the foot?

nemakam:
Makes sense. I was initially thinking about letting them use 500ms operation timeout. But if we are moving to seconds format, then maybe we can have a minimum timeout of 1 second.

SeanFeldman:
> But lets not have a lower limit.

It's got to be greater than zero, no?

src/Microsoft.Azure.ServiceBus/ServiceBusConnectionStringBuilder.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Globalization;
```
nemakam: nit: move inside `namespace`

SeanFeldman:
@0xced once you address https://github.com/Azure/azure-service-bus-dotnet/pull/644#discussion_r266104912, please remove `[WIP]` to indicate it's ready and I'll merge to kick off 3.4.0 release. Thanks 🙂

SeanFeldman:
Thank you.

PR #657, opened 2019-03-06
Return zero for message Size when body is null

### LINKED ISSUE #648: NullReferenceException when accessing message.Size property
## Actual Behavior

When accessing empty message (`message.Body` is null) size property `NullReferenceException` is thrown.

## Expected Behavior

Getting NullReferenceException was pretty unexpected. I would expect to get either 0 size or size of message properties

## Versions  

nuget 3.3.0 / netstandard2

SeanFeldman:
Fixes #648 

The old client did not throw for an empty message. The new client shouldn't either.


SeanFeldman:
@nemakam please review & merge

PR #666, opened 2019-03-16
Mute NuGet licenseUrl warning to unblock contributors using Rider IDE

0xced:
Having GeneratePackageOnBuild=true causes the following error:

> /usr/local/share/dotnet/sdk/2.2.105/Sdks/NuGet.Build.Tasks.Pack/buildCrossTargeting/NuGet.Build.Tasks.Pack.targets(202,5): Error  : The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.

This should be a simple warning but since `TreatWarningsAsErrors` is set to true in Directory.Build.props it becomes an error.

Since we don't need to generate a package on build during development anyway, we set GeneratePackageOnBuild=true when building the project with the build.ps1 script only.

**Update**: this was a symptom of [NU5125](https://github.com/NuGet/Home/issues/7509). Muting it for now.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
+++ b/src/Microsoft.Azure.ServiceBus/Microsoft.Azure.ServiceBus.csproj
@@ -23,7 +23,7 @@
     <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <NoWarn>CS1591;CS1573</NoWarn>
+    <NoWarn>CS1591;CS1573;NU5125</NoWarn>
   </PropertyGroup>
 
   <ItemGroup>


0xced:
The problem is that the replacement of the licenseUrl [is not yet implemented on NuGet](https://github.com/NuGet/Home/issues/7509#issuecomment-445528889)!

How do you work on your development machine? Don't you get the same error as me? I had to delete `<GeneratePackageOnBuild>True</GeneratePackageOnBuild>` from the csproj to be able to work on this project. (I’m using JetBrains Rider on macOS)

SeanFeldman:
It generates an error, yes, but doesn't break the build.

0xced:
OK, so I have taken another path: silencing the `NU5125` warning so that we don't get an error.

By the way: the error prevents Rider to run unit tests.

SeanFeldman:
😀

dlstucki:
I was noticing this same error in the azure-relay-dotnet repo. Isn't there something like PackageLicenseExpression that one can set to MIT in the csproj file? (I'm away from a computer at the moment)

SeanFeldman:
That's it! 🙂

0xced:
The `PackageLicenseExpression` also [looks new](https://github.com/dotnet/docs/pull/9675). I’m worried that it may end up like the new [license file](https://github.com/NuGet/Home/issues/7509#issuecomment-445528889) with **License expressions are not yet supported.** when uploading it to NuGet.org.

SeanFeldman:
I'd say raise an issue about this, link to this PR and we move forward to unblock contributors with Rider.
BTW, Visual Studio doesn't fail on this warning 😉

0xced:
Out of curiosity, why would you want the NuGet package to be generated during development? I can't think of a good reason to generate a package in debug configuration. I see at least two reasons why you would **not** want to generate it:
1. It's not a package you want to deploy to NuGet.org, so it's useless.
2. The debug build is slowed down by the creation of the package.

I can't think of another project where the package is produced for debug builds.

SeanFeldman:
@0xced I understand you don't like that the package is created in debug mode. Would not generalize though. All projects (OSS) I'm involved in generate packages regardless of the build type. I'm refusing to believe that a few milliseconds that package generation is taking is a significant slow down for you.
I might be off on this, but if this was not an issue for other contributors, why is it suddenly a problem? Rider, that I get. The workaround is to help Rider not to fail. It's also a temporary one until NuGet issue is sorted out.

PR #668, opened 2019-03-20
Enable the use of a proxy when running in .NET Core

### LINKED ISSUE #436: HTTP proxy support in .NET Core/Standard
The documentation, [such that it is](https://github.com/Azure/azure-service-bus-dotnet/blob/9b31751a1c5ade61f353bd4905b0665ad0269dfa/src/Microsoft.Azure.ServiceBus/TransportType.cs), for proxy support seems to conflict with [this](https://github.com/dotnet/corefx/pull/11996) .NET Standard PR.

Is using an HTTP proxy supported when using transport `AmqpWebSockets` with .NET Core clients?

I'd like to use ASB from .NET Core clients running on Linux, but lack of HTTP proxy support is a deal breaker for me.

ErikMogensen:
I believe this fixes #436 

It has been been tested running in .NET Framework 4.7.2 and .NET Core 2.1 on Windows. It is assumed that the proxy is only used in transportMode=AmqpWebSockets.

### DIFF
--- a/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
+++ b/src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.Azure.ServiceBus
 {
     using System;
+    using System.Net;
     using System.Threading.Tasks;
     using Microsoft.Azure.Amqp;
     using Microsoft.Azure.Amqp.Framing;
@@ -309,7 +310,8 @@ namespace Microsoft.Azure.ServiceBus
                 return AmqpConnectionHelper.CreateWebSocketTransportSettings(
                     networkHost: networkHost,
                     hostName: hostName,
-                    port: port);
+                    port: port,
+                    proxy: WebRequest.DefaultWebProxy);
             }
 
             return AmqpConnectionHelper.CreateTcpTransportSettings(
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
```diff
@@ -309,7 +310,8 @@ TransportSettings CreateTransportSettings()
                 return AmqpConnectionHelper.CreateWebSocketTransportSettings(
                     networkHost: networkHost,
                     hostName: hostName,
-                    port: port);
+                    port: port,
+                    proxy: WebRequest.DefaultWebProxy);
```
nemakam: It seems weird that this is required. ClientWebSocketOptions should ideally just use the defaultWebProxy. 
Have you been able to get a repro of proxy working after this change, and also made sure it doesn't work on your setup without this change?
In the full framework, it already uses the defaultWebProxy and hence it has been working.

nemakam:
For the sake of others, would you please describe how did you set the default proxy on this app?

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
ErikMogensen: Yes, it does not work behind the proxy without this change running on .NET Core 2.1 on Windows 10. When running in the full Framework it works without this change.

ErikMogensen:
Sure, I wrote about it in the issue #436 , but I'll put it here as well. I used :
```
WebProxy proxy = new WebProxy(
    "http://<proxyserver>:<proxyport>", 
    true));
proxy.UseDefaultCredentials = true;
WebRequest.DefaultWebProxy = proxy;
```
to set the proxy.

src/Microsoft.Azure.ServiceBus/ServiceBusConnection.cs
nemakam: That is interesting. Thanks

PR #671, opened 2019-03-21
Add the possibility to modify properties when abandoning a message

0xced:
This is useful to store exception details when abandoning a message and still keeping the convenience of registering a message or session handler with AutoComplete = true


test/Microsoft.Azure.ServiceBus.UnitTests/Microsoft.Azure.ServiceBus.UnitTests.csproj
```diff
@@ -6,6 +6,7 @@
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <LangVersion>7.3</LangVersion>
```
SeanFeldman: Make it `latest`

0xced:
> Are we giving promises that everytime we run into exception, we would be updating the properties?

Only when the exception happens in the **user callback**, i.e. the `handler` passed to `void IReceiverClient.RegisterMessageHandler(Func<Message, CancellationToken, Task> handler, …)`

> There might be lot of exceptions that are raised for which we will not be able to update them. For example, CommunicationException, or say LockLostException etc.

I’m not sure about `ServiceBusCommunicationException`, but the message is **not** abandoned if the exception is a `MessageLockLostException`:

https://github.com/Azure/azure-service-bus-dotnet/blob/4e6ca73c0f661c944a5685d60deb59bcd2222edc/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs#L153-L157

By the way, I wonder how a `MessageLockLostException` could be thrown from the **user callback**. 🤔

> There are lot of places where we call the exception handler. What is the expectation from the handler code in those cases?

I have documented this: the returned dictionary is ignored. Maybe this documentation needs improvement?
https://github.com/Azure/azure-service-bus-dotnet/blob/4e6ca73c0f661c944a5685d60deb59bcd2222edc/src/Microsoft.Azure.ServiceBus/MessageHandlerOptions.cs#L43-L44

> How will it distinguish between exception thrown from the message handler, vs exception thrown from doing auto complete?

I’m not sure I fully understand your concern. I think it doesn’t matter since the returned dictionary is ignored when `RaiseExceptionReceived` is called when auto completion fails:
https://github.com/Azure/azure-service-bus-dotnet/blob/4e6ca73c0f661c944a5685d60deb59bcd2222edc/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs#L210-L224

If you meant from the user of the library point of view, the `ExceptionReceivedEventArgs` passed to the exception handler has a `ExceptionReceivedContext.Action` which describes the [action](https://github.com/Azure/azure-service-bus-dotnet/blob/4e6ca73c0f661c944a5685d60deb59bcd2222edc/src/Microsoft.Azure.ServiceBus/ExceptionReceivedEventArgsAction.cs), i.e. *Complete*, *Abandon*, *UserCallback*, *Receive*, *RenewLock*, *AcceptMessageSession* or *CloseMessageSession*.

test/Microsoft.Azure.ServiceBus.UnitTests/Microsoft.Azure.ServiceBus.UnitTests.csproj
0xced: I set it explicitly to 7.3 because C# 7.3 is needed for the tests to compile after I added the second `MessageHandlerOptions` constructor. See [Unable to invoke overload when argument is func of non generic task and supplied as method](https://stackoverflow.com/questions/42642355/unable-to-invoke-overload-when-argument-is-func-of-non-generic-task-and-supplied#comment97274139_42642355) which describes the exact same situation where the compiler generates errors if LangVersion is smaller than 7.3:
```
OnSessionQueueTests.cs(63, 47): [CS0407] 'Task OnSessionQueueTests.ExceptionReceivedHandler(ExceptionReceivedEventArgs)' has the wrong return type
OnSessionQueueTests.cs(147, 47): [CS0407] 'Task OnSessionQueueTests.ExceptionReceivedHandler(ExceptionReceivedEventArgs)' has the wrong return type
OnSessionTopicSubscriptionTests.cs(104, 47): [CS0407] 'Task OnSessionTopicSubscriptionTests.ExceptionReceivedHandler(ExceptionReceivedEventArgs)' has the wrong return type
SenderReceiverClientTestBase.cs(262, 43): [CS0407] 'Task SenderReceiverClientTestBase.ExceptionReceivedHandler(ExceptionReceivedEventArgs)' has the wrong return type
SenderReceiverClientTestBase.cs(292, 43): [CS0407] 'Task SenderReceiverClientTestBase.ExceptionReceivedHandler(ExceptionReceivedEventArgs)' has the wrong return type
```

But I’m happy to bump it to `latest` if you think it’s better.

SeanFeldman: Test project is free to use the latest and greatest. Rather than bumping it manually all the time, it could be on the latest all the time.

0xced:
@nemakam Any thoughts on this?

